PR #153, opened 2025-02-19
Viz: Prep for interactive interactions by refactoring the Lir intermediate representation and focusing it on the graph-based representation

ym-han:
The primary goal of this PR is to refactor the intermediate representation used for the UI. 

There used to be two IRs in a way, a tree-shaped one that's quite similar to the input interfaces (e.g., the LirNodes included And/Or/Not) and a graph-based one that was produced on the way to spitting out the final graph. The one that was closer to the input interfaces used to be the primary IR. But  it became clear at one of the meetings last week that (i) the focus for the UI is on graph-based operations --- e.g. getting/setting data for an edge or vertex in the graph --- and (ii) contrary to something I had been thinking and had heard before, we probably won't need to do things that the old tree-shaped And/Or/Not representations would have facilitated.

For this reason, it made sense to refactor the Lir and have the main Lir objects be those corresponding to the graph-based representation --- to have the primary UI IR be one for which there's a bijection between that and the concrete UI.

Sub-tasks:

- [x] Add the base functionality for generating all the paths through the DAG (but without printing it out properly for now)
- [x] Refactor most of the Viz software design / interfaces: remove the old Lir nodes and make the key intermediate Lir representation graph-based. This also allows us to simplify the transformation of IRExprs to the dag.
- [x] Finish the refactor by making something like a `LadderGraphLirNode` that hides the Dag<LirId> and use that instead for the body of the FunDeclLirNode. This `LadderGraphLirNode` should have getters/setters for getting the vertices and edges of the dag, and hence pave the way for doing things like getting/setting data associated with the vertices/edges in the future
- [x] Diagnose and fix issue with SF nodes getting wrong measurements when running in dev mode --- turns out it was because they were sometimes getting rendered and measured in during SSR, in a non-browser environment. (The main reason for looking into this was to make sure there wasn't some big, fundamental mistake with how the layouting is being synchronized with rendering / measuring in the visualizer.)
- [x] Make the handling of the LirId x SF Id more future-proof and more easily debuggable

Re how this IR will be used for stateful itneractions in the future: this IR will be wired up to the SvelteFlow and things like the path displayer UI, so that, e.g., if the user mouses over the UI, that will trigger some functionality in the IR that in turn causes some UI to change. Any mention of state-related things in this PR is very tentative --- that stuff will only really be handled in the next PR.


ts-apps/decision-logic-visualizer/src/lib/algebraic-graphs/dag.ts
```diff
@@ -77,37 +77,60 @@ export abstract class Dag<A extends Ord<A> & HasId>
...
 /** Empty graph */
-export class Empty<A extends Ord<A> & HasId> extends Dag<A> {
+export class Empty<A extends Ord<A>> extends Dag<A> {
   constructor() {
     super()
   }
 }
 
-export function isVertex<A extends Ord<A> & HasId>(g: DirectedAcyclicGraph<A>) {
+export function isVertex<A extends Ord<A>>(g: DirectedAcyclicGraph<A>) {
```
fendor, line 124:
```suggestion
export function isVertex<A extends Ord<A>>(g: DirectedAcyclicGraph<A>): g is Vertex<A> {
```
Then it becomes a proof that allows you to narrow down the type of `g`.

ts-apps/decision-logic-visualizer/src/lib/data/viz-expr-to-lir.ts
```diff
@@ -26,38 +30,116 @@ export const VizDeclLirSource: LirSource<IRDecl, DeclLirNode> = {
...
+* Preconditions / assumptions:
+* - To reduce visual clutter in the generated graph, the input ANDs/ORs should
+    have already been 'flattened'; i.e.,
+    nested exprs like (AND [a (AND [b])]) should have already been flattened to (AND [a b]).
+*
+*/
+export const LadderGraphLirSource: LirSource<IRExpr, LadderGraphLirNode> = {
+  toLir(nodeInfo: LirNodeInfo, expr: IRExpr): LadderGraphLirNode {
+    const overallSource = vertex(
+      new SourceLirNode(nodeInfo).getId()
```
fendor, line 49: It might be a little bit more direct, if `LirId` had a static `getFreshId()`  or something like that, than you could avoid the indirection.
Just an idea, though.

ts-apps/decision-logic-visualizer/src/lib/data/viz-expr-to-lir.ts
```diff
@@ -26,38 +30,116 @@ export const VizDeclLirSource: LirSource<IRDecl, DeclLirNode> = {
...
+  return match(expr)
+    .with({ $type: 'BoolVar' }, (originalVar) => {
+      const boolvar = new BoolVarLirNode(nodeInfo, originalVar)
+      return vertex(boolvar.getId())
+    })
+    .with({ $type: 'Not' }, (neg) => {
+      const negand = transform(nodeInfo, neg.negand)
+      const notStart = vertex(
+        new NotStartLirNode(nodeInfo).getId()
+      ) as DirectedAcyclicGraph<LirId>
```
fendor, line 105: Unnecessary cast again?

ts-apps/decision-logic-visualizer/src/lib/data/viz-expr-to-lir.ts
```diff
@@ -26,38 +30,116 @@ export const VizDeclLirSource: LirSource<IRDecl, DeclLirNode> = {
...
+      const notEnd = vertex(
+        new NotEndLirNode(nodeInfo).getId()
+      ) as DirectedAcyclicGraph<LirId>
```
fendor, line 108: Cast

ts-apps/decision-logic-visualizer/src/lib/data/viz-expr-to-lir.ts
```diff
@@ -26,38 +30,116 @@ export const VizDeclLirSource: LirSource<IRDecl, DeclLirNode> = {
...
+    })
+    .with({ $type: 'Or' }, (orExpr) => {
+      const children = orExpr.args.map((n) => transform(nodeInfo, n))
+
+      const overallSource = vertex(
+        new SourceLirNode(nodeInfo).getId()
+      ) as DirectedAcyclicGraph<LirId>
+      const overallSink = vertex(
+        new SinkLirNode(nodeInfo).getId()
+      ) as DirectedAcyclicGraph<LirId>
```
fendor, line 131: Casts and Ids

ts-apps/decision-logic-visualizer/src/lib/algebraic-graphs/dag.ts
ym-han: Great catch, thank you!

ts-apps/decision-logic-visualizer/src/lib/data/viz-expr-to-lir.ts
```diff
@@ -26,38 +30,116 @@ export const VizDeclLirSource: LirSource<IRDecl, DeclLirNode> = {
...
+* - To reduce visual clutter in the generated graph, the input ANDs/ORs should
+    have already been 'flattened'; i.e.,
+    nested exprs like (AND [a (AND [b])]) should have already been flattened to (AND [a b]).
+*
+*/
+export const LadderGraphLirSource: LirSource<IRExpr, LadderGraphLirNode> = {
+  toLir(nodeInfo: LirNodeInfo, expr: IRExpr): LadderGraphLirNode {
+    const overallSource = vertex(
+      new SourceLirNode(nodeInfo).getId()
+    ) as DirectedAcyclicGraph<LirId>
```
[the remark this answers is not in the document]
ym-han: Nice catch again! The type assertion might have been needed at some point, but you're right, it's not necessary now.

ts-apps/decision-logic-visualizer/src/lib/data/viz-expr-to-lir.ts
ym-han: Ah what's happening here isn't that we are making a new id, but rather that we are making a new LirNode and then getting its id.

ym-han:
Thanks for pointing out the unnecessary casts / type assertions --- I've removed them (and a couple other ones that I found)!

Sidenote: I've also been trying to make the PRs smaller. (This one could actually have been split into at least two PRs as well, but at least I managed to stop myself from squeezing in even more.) If you have any feedback, that'd be helpful --- e.g. 'PRs of this size are more manageable, but in general the smaller the better', or 'This is a step in the right direction, but it's still quite a bit bigger than is ideal'.

PR #161, opened 2025-02-20
standalone WebApp

MangoIV:
give a stab at implementing a webapp for the vscode extension

Testing the waters of how much work it would be to get it running in the web. Current problems 

- [x] haskell lsp implementation does not know how to deal with `inmemory://` uris
- [x] get semantic highlighting to work
- [x] make the server implementation more robust currently just a thread that processes incoming and outgoing websocket requests 
- [x] deployment
- [x] try to show the webapp (currently only the editor with language client works) 
- [x] audit sandboxing of lsp server
- [x] clean up haskell code


ym-han:
I was just looking at the recording of the meeting. Re the 'global state' you talked about for the webview, that is **VSCode-specific**: I don't think you need that (or should use that) with Monaco.

I don't have actual experience with Monaco, but just from skimming some docs /  tutorials, it looks like it should be a lot simpler than integrating with VSCode. (I.e., it's probably not a good idea to literally take the current webview, which has been tailored to work specifically with VSCode, and just try to integrate it wholesale.) I would need to look at the Monaco docs / examples more to be sure, but I think integrating the visualizer would be not be fundamentally different from integrating any other kind of UI component into a web app. One would basically want to write something that's similar in spirit but simpler than the current webview. There wouldn't be a need for the VSCode messenger, for example.

And for the client x server communication, I think you can use methods like `onNotification` or `onRequest` as shown in https://langium.org/docs/learn/minilogo/generation_in_the_web/ (though again I would have to look at it more to be sure)

ym-han:
@MangoIV small thing, but if you are using SSR, you'll want to make sure the visualizer is set up in such a way so that the visualizer part does *not* undergo SSR. This is because if SvelteFlow nodes can render on the server, one can run into a situation where they render on the server and get incorrect measurements, incorrect measurements that then result in weird-looking / out-of-sync visuals.

jl4/lsp/LSP/L4/Rules.hs
MangoIV, line 1: need to fix this, I accidentally reformatted the thing

jl4/src/L4/Parser/SrcSpan.hs
MangoIV, line 1: restore

jl4/src/L4/TypeCheck.hs
MangoIV, line 1: restore

nix/hetzner.nix
```diff
@@ -0,0 +1,38 @@
...
+    "ahci"
+    "xhci_pci"
+    "virtio_pci"
+    "virtio_scsi"
+    "sd_mod"
+    "sr_mod"
+  ];
+  boot.initrd.kernelModules = [ ];
+  boot.kernelModules = [ ];
+  boot.extraModulePackages = [ ];
```
MangoIV, line 33: remove

ts-apps/jl4-web/src/routes/+layout.svelte
```diff
@@ -0,0 +1,6 @@
+<script>
+  import '../style.css';
+  import '../../../decision-logic-visualizer/src/style.css'
```
ym-han, line 3: It's not a good idea to import the CSS from the dlv package directly like this. It's better to do it in the style of

https://github.com/smucclaw/l4-ide/blob/cb9aa6a430d3357b46cd4962416508b6711ea858/ts-apps/webview/src/app.css#L5

where we 

(i) import it from [the export path that's declared in the dlv package.json](https://github.com/smucclaw/l4-ide/blob/cb9aa6a430d3357b46cd4962416508b6711ea858/ts-apps/decision-logic-visualizer/package.json#L29) and 

(ii) use the path alias supplied by turborepo.

(It probably doesn't matter whether you import it in the script or as an `@import` in the app.css, but deviating from (i) and (ii) may lead to unexpected or undefined behavior.)

ts-apps/jl4-web/tailwind.config.js
```diff
@@ -0,0 +1,16 @@
+/** @type {import('tailwindcss').Config} */
```
ym-han, line 1: btw you shouldn't need to use tailwind in jl4-web in order to use the dlv lib. (e.g., the `webview` package does not use tailwind.) If you find yourself *needing* to do it to make the CSS work, that's a red flag.

(There might also some other hoops that have to be jumped through when (i) you want to use tailwind and (ii) also use a library that itself uses tailwind, and I don't see that being done here. I don't personally know Tailwind / JS infra well enough to be able to figure out that stuff quickly; I would recommend just not using tailwind for jl4-web unless you really really have to (and right now I don't see any real use of it).

ts-apps/jl4-web/src/routes/+page.svelte
```diff
@@ -0,0 +1,247 @@
...
+      "vscode-languageclient/browser.js"
+    );
+    const { MonacoLanguageClient } = await import("monaco-languageclient");
+    const { WebSocketMessageReader, WebSocketMessageWriter, toSocket } =
+      await import("vscode-ws-jsonrpc");
+    const { configureDefaultWorkerFactory } = await import(
+      "monaco-editor-wrapper/workers/workerLoaders"
+    );
+    const { ConsoleLogger } = await import("monaco-languageclient/tools");
+    const { Schema } = await import("effect");
```
ym-han, line 60: How come `const { Schema } = await import("effect");` needs to be imported like this?

ym-han (changes requested):
The main thing I would really prefer that we address before merging is the CSS infra: the importing and the Tailwind in the jl4-web.

(I take it there isn't a real rush anyway to merge to main given that you've managed to deploy from this branch?)

ts-apps/jl4-web/src/routes/+page.svelte
MangoIV: The typescript error doesn't go away if we don't import it that way but I can try whether it works (while ignoring the type error) and change it.

ym-han: I can experiment with that, I just wanted to try to understand the intent so I don't inadvertently break something

ts-apps/jl4-web/tailwind.config.js
MangoIV: so the preprocessing of the imported tailwind cdd files will be automatically done by the other repo? I wouldn't have expected that. But I can try without this.

ts-apps/jl4-web/src/routes/+page.svelte
MangoIV: @ym-han btw if you have any idea of how to solve this type error or make it simpler s.t. it can't show a type error, that would be great.

ts-apps/jl4-web/src/routes/+layout.svelte
ym-han: I'm happy to try changing it -- my personal inclination would be to remove Tailwind (as noted in https://github.com/smucclaw/l4-ide/pull/161#discussion_r1972933227) unless you have an independent reason for wanting it

MangoIV: yes - less is better, if you manage to remove it that would be awesome!

ts-apps/jl4-web/src/routes/+page.svelte
ym-han: > @ym-han btw if you have any idea of how to solve this type error or make it simpler s.t. it can't show a type error, that would be great. (because I personally cannot make sense of the error)

I'll experiment with it, yeah. But it would helpful if you could explain the intent behind those await imports in the first place so that I don't accidentally break something or waste too much time going down the wrong path. I don't have that much frontend experience.

ts-apps/jl4-web/src/routes/+layout.svelte
ym-han: I'll wrok on it!

ts-apps/jl4-web/src/routes/+page.svelte
ym-han: Yep!

ts-apps/jl4-web/tailwind.config.js
ym-han: I can do it

fendor (approved):
LGTM, I think websocket stuff shouldn't be part of the language server, but otherwise, this is amazing!
Great work here @MangoIV, you discovered quite a number of improvements for the language server which we will upstream to HLS itself as well. :confetti_ball:

ts-apps/jl4-web/src/routes/+layout.svelte
ym-han: (I'll also look into the export of the css by the `decision-logic-visualizer` package --- I might need to do more to get the styles correctly exported.)

MangoIV:
@ym-han  Lovely. I think the only thing that is a bit worse than before is that the background of the rhs column was pre styled by tailwind and it's now black. I think we should fix that before merging.

ym-han:
> @ym-han Lovely. I think the only thing that is a bit worse than before is that the background of the rhs column was pre styled by tailwind and it's now black. I think we should fix that before merging.

Yep I was just thinking about that — I should be able to fix that by tonight/early morning  my time

ym-han:
@MangoIV I've fixed the type error with the Effect stuff, but am running into another issue. I'm too sleepy now to be able to debug it, so I've opened a PR that details my progress so far, in case you want to get this done asap: https://github.com/smucclaw/l4-ide/pull/186 I can also continue working on it tomorrow.

MangoIV:
@ym-han I will shortly merge all the changes to the haskell code and then I will probably squash this entire branch into changes to the typescript code and the nix code (which also is ready but depends on the typescript code)

I think then we'll have a bit less pressure to complete fixing the typescript build.

ym-han:
> @ym-han I will shortly merge all the changes to the haskell code and then I will probably squash this entire branch into changes to the typescript code and the nix code (which also is ready but depends on the typescript code)
> 
> I think then we'll have a bit less pressure to complete fixing the typescript build.

Sounds good, thank you!

MangoIV:
@ym-han did we already resolve the issue wrt the vscode extension not building?

MangoIV:
> I can look at that next week.

perfect, thank you, I think merging after that has been done is a good idea.

MangoIV:
@ym-han it's not that urgent anymore, since I unblocked us on the haskell code.

ym-han:
> @ym-han it's not that urgent anymore, since I unblocked us on the haskell code.

Ah that's very helpful to know, thank you.

I was going to say though that I would probably need at least 3 - 4 days to fix this. Based on what I've seen of this so far (and based on my experience so far with this sort of thing), I don't think I can really make meaningful progress without investing some time upfront studying the relevant things (package managers, bundling, dep management esp. in a monorepo setting, etc) more systematically and in more detail.

ym-han:
Btw if in the future we need to merge viz/frontend-related things from `main`, feel free to leave that to me (esp. if there are merge conflicts or anything like that)

MangoIV:
a thorough look at the package-lock.json showed a suspicious section and the suspicion was correct. this was it, it's fixed now.

ym-han:
> a thorough look at the package-lock.json showed a suspicious section and the suspicion was correct. this was it, it's fixed now.

Very, very nice --- thank you for fixing this, @MangoIV !!!!!!

I'm a bit worried that the configuration of the monorepo / projects might not be doing enough to just prevent this sort of issue from arising in the first place. (E.g. I had tried to add svelte to the root-level `overrides` of the package.json, but that wasn't enough to fix the issue, and it feels a priori odd that one has to manually change the package-lock.json -- though maybe this is just how things are in JS land.) I'll try to look more into this after I've finished the next milestone for the visualizer.

PR #164, opened 2025-02-21
Viz: Implement simple bool var node value change interaction pattern. Add support for edge attributes in the underlying graph structures.

ym-han:
## The key things that this PR does

Implements the following bool var node value change interaction pattern:
* User can click on each bool var node to indicate whether it's true or false (or they don't know). Values cycle with clicks.
* Node styling updates with changes in the bool var node value.
* Co-referring (in the sense of pointing to the same Unique) bool var nodes will have their values change in lockstep. (This is still quite crude and naive --- it's just using the Uniques from the scope/typechecker in the most naive way.)

Adds support for edge attributes (edge styles and labels) in the underlying graph structures, along with some tests.
* This is a step towards implementing things like the interaction pattern for updating edge styles and explanatory edge labels.
* But those things will only be fully implemented in subsequent PRs.
* **I.e., please don't bother reviewing the edge attributes stuff in this PR -- that gets improved in the next PR.**

The Not nodes will probably need to be modified to get a UI/UX that meshes better with the node value state visuals, but I'm deferring that for now.

## Example screenshots

If you click on one of the `p`s once, you get

image

If you click twice (note: the background colors have been made a little lighter than in the following screenshots)

image

and if you click three times, we're back where we started.
image

----

### Edit: The background colors have been made lighter

image

## More fine-grained subtasks

- [x] Add pub/sub to LirRegistry / copy/adapt it from `hoare-logic-tutor` project
- [x] Add support for edge attributes (edge styles and labels) in the underlying graph structures, along with some tests
- [x] Add a custom SvelteFlow edge --- a Ladder Edge --- that supports edge attributes like a label and the stroke color
- [x] Fix the name of the CSS class for stroke in style.css
- [x] Update Ladder Lir Nodes with the positions and measured dimensions of the sf nodes after the nodes are layouted
- [x] flow-base.svelte: Implement the simplest possible `onLadderGraphNonPositionalChange` that re-assigns the SF `NODES` and `EDGES` when the LadderLirNodes publish a change in a non-position attribute
- [x] Implement UI for letting users change the values on the nodes
- [x] Add an Environment object with mappings Unique -> Value and Unique -> Set of LirId
- [x] Update the LirNodes with positional info from the SF Nodes on drag stop
- [x] Add docs about which pieces of state the LirNodes own and do not own
- [x] Improve the information hiding of `lirIdToSFId`
- [x] Add a `lirRegistry` prop to LadderFlow and pass in the prop in uses of the library
- [x] @meowyam helped with thinking through the ui ux and the point of being able to toggle values
- [x] @johsi-k helped with studying and analyzing both Meng's past comments on slack and Meng's `layman` visualizer, and trying to understand what the intended interaction paradigm of that visualizer is supposed to be. She also helped with discussing what the ideal ui ux should be.

----

Misc:

- [x] Refactor: Improve the checking of whether the NODES have been measured in doLayout
- [x] Refactor: Improve viz container css
- [x] Deps: Update xyflow svelte dep to get useful bug fixes


MangoIV (approved):
lgtm, nice!

Issue #44, opened 2025-01-31
jl4: Add an AKA construct to the language


kosmikus:
Idea:

Everywhere where we are introducing a new name, we can also use `AKA names{,}+` to introduce additional names for the same construct.

These can be used for grammatical beauty, but also for disambiguation if needed.

Example:
```
DECLARE Person AKA Persons
  HAS name IS A STRING
      age IS A NUMBER

GIVEN ps IS A LIST OF Persons -- previously not accepted
...
```

PR #172, opened 2025-02-25
Add type synonyms

kosmikus:


kosmikus:
This is a part of #44 .

jl4/src/L4/Parser.hs
```diff
@@ -362,26 +362,37 @@ declare sig =
 
 typeDecl :: Parser (TypeDecl Name)
 typeDecl =
-  recordDecl <|> enumDecl
+  recordDecl <|> enumOrSynonymDecl
```
MangoIV, line 365: nit: I would prefer if this were
`recordDecl <|> enumDecl <|> synonymDecl`

jl4/src/L4/TypeCheck.hs
```diff
@@ -582,10 +652,38 @@ addError e = do
   ctx <- use #errorContext
   with (MkCheckErrorWithContext e ctx)
 
+-- | Should never return 'Nothing' if our system is OK.
+getEntityInfo :: Resolved -> Check (Maybe CheckEntity)
+getEntityInfo r = do
+  ei <- use #entityInfo
+  pure (snd <$> Map.lookup (getUnique r) ei)
```
MangoIV, line 659: why not return `Check CheckEntity` then - I have seen this used in another place where this would return `Nothing` for that function if this were a `Nothing`, I think I'd rather have this explode here if we had a bug than silently keep on going and potentially returning wrong results from functions that use this?

jl4/src/L4/TypeCheck.hs
```diff
@@ -1696,7 +1808,41 @@ unify (InfVar _ann1 _pre1 i1) t2
...
+  case mt1' of
+    Just t1' -> unify t1' t2
+    Nothing  -> do
+      mt2' <- tryExpand t2
+      case mt2' of
+        Just t2' -> unify t1 t2'
+        Nothing  -> unifyBase t1 t2
+  where
+    tryExpand (TyApp _ann n ts) = tryExpandTypeSynonym n ts
+    tryExpand _                 = pure Nothing
```
MangoIV, line 1828: you could inline `tryExpandTypeSynonym` into `tryExpand` and you could use the `Alternative` instance of `MaybeT Check` to improve readability a bit.

jl4/src/L4/TypeCheck.hs
```diff
@@ -1696,7 +1808,41 @@ unify (InfVar _ann1 _pre1 i1) t2
...
+unifyBase :: Type' Resolved -> Type' Resolved -> Check Bool
+unifyBase (TyApp _ann1 n1 ts1) (TyApp _ann2 n2 ts2) = do
+  -- both are type constructors or type variables
+  r <- ensureSameRef n1 n2
+  -- We should not need to check the same length because we've done kind checking.
+  rs <- traverse (uncurry unify) (zip ts1 ts2)
+  pure (and (r : rs))
+unifyBase (Fun _ann1 onts1 t1) (Fun _ann2 onts2 t2)
+  | length onts1 == length onts2 = do
+    traverse_ (uncurry unify) (zip (optionallyNamedTypeType <$> onts1) (optionallyNamedTypeType <$> onts2))
```
MangoIV, line 1842: why does this ignore the result of the unification?

jl4/src/L4/Parser.hs
kosmikus: This is an artifact of LL(1) essentially. Both `enum` and `synonym` decls start with `IS`, and if you want to avoid unnecessary uses of `try`, you have to group them.

jl4/src/L4/TypeCheck.hs
kosmikus: Thank you, that's a real bug.

jl4/src/L4/TypeCheck.hs
kosmikus: I can try. I'm not convinced it will be much more readable though.

jl4/src/L4/TypeCheck.hs
kosmikus: I'm really reluctant to introduce places where the server explodes as long as that makes the editor / language server integration crash weirdly. I'd be happy to throw an internal error *and* continue, but in order to continue, I'd still have to produce `Nothing` then, because there's no good default value.

jl4/src/L4/TypeCheck.hs
MangoIV: > (But one introduced earlier.)

yeah I'm guessing from when you changed `unify` to return `Bool` instead of `()`

kosmikus: Fixed in f91235d63c1515d3e95c0f1fdc6b42540766857b.

jl4/src/L4/TypeCheck.hs
kosmikus: "Fixed" in d07e53bdfca651f35ad658e73e74ca11005cc788.

jl4/src/L4/TypeCheck.hs
kosmikus: Added an internal error in 961f98c645ec4e37c344a1178c45704515291563.

PR #183, opened 2025-02-27
Implement AKA

### LINKED ISSUE #44: jl4: Add an AKA construct to the language

kosmikus:
Fixes #44 .

### DIFF
--- a/jl4/src/L4/Syntax.hs
+++ b/jl4/src/L4/Syntax.hs
@@ -115,7 +115,12 @@ data GivethSig n =
   deriving anyclass (SOP.Generic, ToExpr, NFData)
 
 data Decide n =
-  MkDecide Anno (TypeSig n) (AppForm n) (Expr n)
+  MkDecide Anno (TypeSig n) (AppFormAka n) (Expr n)
+  deriving stock (GHC.Generic, Eq, Show)
+  deriving anyclass (SOP.Generic, ToExpr, NFData)
+
+data AppFormAka n =
+  MkAppFormAka Anno (AppForm n) (Maybe (Aka n))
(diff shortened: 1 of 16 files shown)


jl4/src/L4/Syntax.hs
```diff
@@ -115,7 +115,12 @@ data GivethSig n =
   deriving anyclass (SOP.Generic, ToExpr, NFData)
 
 data Decide n =
-  MkDecide Anno (TypeSig n) (AppForm n) (Expr n)
+  MkDecide Anno (TypeSig n) (AppFormAka n) (Expr n)
+  deriving stock (GHC.Generic, Eq, Show)
+  deriving anyclass (SOP.Generic, ToExpr, NFData)
+
+data AppFormAka n =
+  MkAppFormAka Anno (AppForm n) (Maybe (Aka n))
```
MangoIV, line 123: what is the reason for not merging AppForm and AppFormAka? Are there constructs that have an AppForm but no possibility to add an aka?

MangoIV (approved):
looks good to me

jl4/src/L4/Syntax.hs
kosmikus: This is a good idea. I'll try to do it.

PR #184, opened 2025-02-27
Set the div immediately wrapping SvelteFlow to height:100%, so that container divs have more flexibility to constrain the height of the flow. This also removes hidden coupling wrt choice of height between webview css and the css in flow-base.svelte

ym-han:


MangoIV:
@fendor may want to have a look as well?

ts-apps/webview/src/routes/+page.svelte
```diff
@@ -98,6 +100,11 @@
     }
   }
 
+  /** So there's space for the fn name */
+  .slightly-shorter-than-full-viewport-height {
+    height: 96svh;
+  }
```
MangoIV, line 106: is that how you do it? shouldn't the fn name take its space and then what comes below fill the remaining space?

MangoIV (approved):

ts-apps/webview/src/routes/+page.svelte
ym-han: The fn name does take its space, you're right. But if I recall correctly, if the height of the div containing the flow is set to some svh, call it *x*, then the height of that div will still be *x* regardless of the height of the fn name element. It's just that the place where the div containing the flow begins that will change when the height of the fn name changes.

So the thought is, if one wants everything to fit nicely on first load in the vscode webview, as I do, then it makes sense to make height of the div containing the flow a bit smaller.

I was experimenting quite fast with this just now and I definitely don't have all the details of the various css height rendering algorithms down, so I might very well be wrong about the details.

PR #186, opened 2025-02-27
Stuck webapp

ym-han:
I've fixed the type error type error with the Effect stuff (was due to mismatches in the versions of Effect used in viz-expr vs jl4-web), but I'm now running into another issue. 

```
Uncaught Error: Default api is not ready yet, do not forget to import 'vscode/localExtensionHost' 
and wait for services initialization
```

I get this issue even if I use 
```
const { makeVizDecoder } = await import("@repo/viz-expr");
```
instead.

I'm opening this just in case you want to fix this asap


MangoIV:
@ym-han thank you for investigating! Take your time, don't feel urged too much :)

MangoIV:
@ym-han maybe you can try to revert everything except your change that fixes the `effect` version?

ym-han:
> @ym-han maybe you can try to revert everything except your change that fixes the `effect` version?

Yeah, thanks for the reminder, I've just filed a PR that just implements that fix. I was getting a bit too sidetracked trying to I understand why I had been running into the other monaco issues I'd run into (even with onMount etc), but maybe it was just some weird issue with stale caches somewhere.

PR #188, opened 2025-02-28
Split the type checker into several modules

kosmikus:
Builds on #183


MangoIV:
image
you're right, it's still the main module that is taking up all the time.

MangoIV (approved):
looks good to me, although it sadly appears that it didn't result in a build time improvement

PR #191, opened 2025-02-28
Fix Effect-related type error (was due to conflicting dependency versions)

ym-han:


MangoIV:
Something went wrong with the git history. It’s hard to review like this.

ym-han:
> Something went wrong with the git history. It’s hard to review like this.

Sorry about that, I'll fix it

PR #194, opened 2025-03-03
Move ci to Docker

kharus:


.github/workflows/build.yml
```diff
@@ -16,13 +16,37 @@ jobs:
...
+          libgmp-dev
+          libgmp10
+          libncurses-dev
+          pkg-config
+          zlib1g-dev
+      - name: Update git config
+        run: git config --global --add safe.directory $GITHUB_WORKSPACE
       # FIXME: this is a temporary fix, the docker container should never be dirty
       # at the begin of a run
       - name: Remove cabal store
```
MangoIV, line 52: i think this should be removed again, if this works now.

MangoIV (approved):
LGTM

PR #205, opened 2025-03-05
Add decision web-service which allows you to run L4 functions through a REST API

fendor:
This ports the web-service from the smucclaw/dsl repository.
As such, it contains a couple of design decision which made sense for
the dsl repo, but do not make a lot of sense right now.

For now, I deliberately decided to keep old code around, for the
non-zero chance that we might need this code again.

In short, this adds a REST service where you can add scenarios/functions
which can later be evaluated using REST endpoints.
As Jl4 does not support some of the features simala or gml do support,
we have some arbitrary restrictions:

* No floating point numbers
* No objects
* No uncertain
* No unknown

There will also be no reasoning tree, as Jl4 doesn't have the tracing
feature of simala, nor the explanation part of gml.

Additionally, the Jl4 endpoint can not handle missing values, and will
always throw an error when trying to invoke a function without all
required values.

We additionally ship a swagger-ui which is served under:

* GET /swagger-ui
* GET /swagger.json

--- 

Adds some semi-related refactorings, since I wanted to reuse the `parseAndCheck` function and disliked that the Lexer doesn't return `PError`. 
So, please review the commits individually.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,6 +45,8 @@ jobs:
           libncurses-dev
           pkg-config
           zlib1g-dev
+          liblzma-dev
+          xz-utils
       - name: Update git config
         run: git config --global --add safe.directory $GITHUB_WORKSPACE
       # FIXME: this is a temporary fix, the docker container should never be dirty
--- a/cabal.project
+++ b/cabal.project
@@ -1,3 +1,4 @@
 packages: 
   ./jl4
+  ./jl4-decision-service
   ./jl4-websessions
--- a/jl4-decision-service/jl4-decision-service.cabal
+++ b/jl4-decision-service/jl4-decision-service.cabal
@@ -1,0 +1,131 @@
+cabal-version: 3.0
+name: jl4-decision-service
+version: 0.1.0.0
+build-type: Simple
+tested-with: ghc ==9.6.6
+extra-source-files:
+  CHANGELOG.md
+
+source-repository head
+  type: git
+  location: https://github.com/cclaw/l4-ide/jl4-decision-service
+
+common defaults
+  default-language: GHC2021
+  ghc-options:
+    -Wall
+    -Wderiving-typeable
+    -Wunused-packages
+    -Werror
+    -Wall
+    -Widentities
+    -Wincomplete-record-updates
+    -Wincomplete-uni-patterns
+    -Wmissing-export-lists
+    -Wmissing-home-modules
+    -Wpartial-fields
+    -Wredundant-constraints
+
+  default-extensions:
+    BlockArguments
+    DefaultSignatures
+    DeriveAnyClass
+    DerivingStrategies
+    DerivingVia
+    DuplicateRecordFields
+    LambdaCase
+    NoFieldSelectors
+    OverloadedLabels
+    OverloadedRecordDot
+    OverloadedStrings
+    -- This is annoying with fourmolu
+    NoImportQualifiedPost
+
+  build-depends:
+    base >=4.7 && <5
+
+library
+  import: defaults
+  exposed-modules:
+    Application
+    Backend.Api
+    Backend.Jl4
+    Examples
+    Options
+    Schema
+    Server
+
+  other-modules:
+    Paths_jl4_decision_service
+
+  autogen-modules:
+    Paths_jl4_decision_service
+
+  hs-source-dirs:
+    src
+
+  build-depends:
+    aeson,
+    aeson-combinators,
+    bytestring,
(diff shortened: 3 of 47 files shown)


MangoIV:
@fendor, I will push directly to this branch.

cabal.project
```diff
@@ -1 +1,2 @@
 packages: ./jl4
+          ./web-service
```
MangoIV, line 2: since I'm also writing a web service, please make this name more speaking. I currently have 
`jl4-websessions`.

fendor: Is it better now? :laughing:

MangoIV: much appreciate it

jl4-decision-service/jl4-decision-service.cabal
```diff
@@ -0,0 +1,131 @@
...
+    -Wunused-packages
+    -Werror
+    -Wall
+    -Widentities
+    -Wincomplete-record-updates
+    -Wincomplete-uni-patterns
+    -Wmissing-export-lists
+    -Wmissing-home-modules
+    -Wpartial-fields
+    -Wredundant-constraints
```
MangoIV, line 27: maybe also add `-Wunused-packages`

jl4-decision-service/src/Backend/Jl4.hs
```diff
@@ -0,0 +1,171 @@
...
+  Eval.ValAssumed var ->
+    throwE $ InterpreterError $ "#EVAL produced ASSUME: " <> prettyLayout var
+
+-- ----------------------------------------------------------------------------
+-- L4 syntax builders
+-- ----------------------------------------------------------------------------
+
+mkTopDeckDirective :: Directive n -> TopDecl n
+mkTopDeckDirective = Directive emptyAnno
+
```
MangoIV, line 113: typo

jl4-decision-service/src/Examples.hs
MangoIV, line 1: I'm guessing these is the only example that we can work with and that's why it's hardcoded?

jl4-decision-service/src/Server.hs
MangoIV, line 1: I've not reviewed this thoroughly

jl4-decision-service/src/Server.hs
```diff
@@ -0,0 +1,854 @@
...
+import GHC.TypeLits
+import Servant
+import System.Timeout (timeout)
+
+-- ----------------------------------------------------------------------------
+-- Servant API
+-- ----------------------------------------------------------------------------
+
+data DbState = DbState
+  { functionDatabase :: TVar (Map Text ValidatedFunction)
```
MangoIV, line 98: I think this could be an MVar.

MangoIV (commented):
I think this looks okay

.github/workflows/build.yml
```diff
@@ -45,6 +45,7 @@ jobs:
           libncurses-dev
           pkg-config
           zlib1g-dev
+          lzma-dev
```
MangoIV, line 48: we could consider using the nix shell here, actually

fendor: The disadvantage is if it breaks, you are likely the only one who could fix it

jl4-decision-service/jl4-decision-service.cabal
fendor: It is on line `19`

jl4-decision-service/src/Examples.hs
fendor: These are builtin examples, but you can add more via the REST API.
They need to be hardcoded somewhere, I am open to better locations.

jl4-decision-service/src/Server.hs
fendor: Could be, but why should it be? You mean less overhead?

.github/workflows/build.yml
MangoIV: you can still go back to doing it "normally"

jl4-decision-service/src/Server.hs
MangoIV: it's simpler if we don't actually need any of the transaction semantics.

MangoIV: just a nit though, can just ignore it

mengwong:
Thanks for the work. Coincidentally, around the same time, I was just writing a motivation for more sophisticated Maybe types in our UI patterns.

https://github.com/smucclaw/l4-ide/blob/mengwong/alcohol/doc/default-logic-2.md

mengwong:
doc followup at #218

PR #206, opened 2025-03-05
Add a way to list the linearized paths through the ladder graph, in a collapsible, scrollable panel. Highlight corresponding path in the ladder graph when mousing over linearized path.

ym-han:
image

image

image

Did not bother highlighting the border of the vertices, since the highlight effect seems clear enough

Tasks:

- [x] Improve edge attrs internal code
- [x] Improve the paths list CSS code (e.g. prob remove all the flex stuff) and code structure somewhat (though more remains to be done here)
- [x] add a 'List paths' button and tuck paths list in a collapsible panel
- [x] do fit view on flow whenever the panel expands or collapses
- [x] Horizontally center the paths in the paths list
- [x] Improve the aesthetics / edge styles of the highlight effect
- [x] Factor paths list container out into another component

For next time:
- Refactor the synchronization of the fit view with the onopenchange to not use a setTimeOut
- factor collapsible out into its own componetn, as suggested by bits ui docs
- improve the styling of the paths list and the list paths button
- the default linearization / pretty printing of the paths in the paths list will be improved (will add 'and', 'or' etc) when I add edge labels
- Might want to make the list paths button smaller
- Might want to make font size for the paths smaller

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3757,6 +3757,34 @@
         "node": "^18.18.0 || ^20.9.0 || >=21.1.0"
       }
     },
+    "node_modules/@floating-ui/core": {
+      "version": "1.6.9",
+      "resolved": "https://registry.npmjs.org/@floating-ui/core/-/core-1.6.9.tgz",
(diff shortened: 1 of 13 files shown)


MangoIV (approved):

ym-han:
Thank you @MangoIV !

PR #209, opened 2025-03-06
Documenting my treatment of regulative rules

mengwong:


MangoIV:
```
npm run format
``` 
should make the CI go green again :)

MangoIV:
I think I need to do a bit more reading about the contract part but I will talk to Andres about that tomorrow.

mengwong:
Matt has offered further commentary and examples that can be used to exercise the theory presented in this PR – see the two doc files on Slack.

mengwong:
fixing formatting, thanks

doc/default-logic-2.md
```diff
@@ -0,0 +1,213 @@
...
+The moment the "motorcycle" node turns `Just True` the overall
+computation evaluates to `True`.
+
+See also https://en.wikipedia.org/wiki/Ternary_logic
+
+## Negation as Failure
+
+The evaluation engine may use "negation as failure" to collapse unknowns to false. `True AND Unknown` is `False`.
+
+## With Defaults
```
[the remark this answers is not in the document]
mengwong: Yes, I am introducing a bunch of ideas here, some of which are orthogonal

mengwong:
Just to be clear, some of this documentation is intended for the legal engineer who is trying to understand L4 syntax and semantics; some of it is intended for L4 internal dev.

doc/regulative.md
```diff
@@ -0,0 +1,451 @@
...
+
+```
+    §  clause 1
+   IF  preconditions
+ UPON  trigger event  -- only used at top-level, otherwise this clause follows from some other clause
+PARTY  p
+   DO  action
+       with  certain criteria
+	   to    some target
+HENCE  clause 2 (... AND clause 4 AND clause 6 OR clause 8)
```
ym-han, line 327: As noted in https://github.com/smucclaw/lam4/blob/88b52216a48a14bcd819aab953da59241b2dffa7/docs/proposal_normative_constructs_aug26_2024.md?plain=1#L218
I'm not sure that the HENCE ... LEST would allow us to express certain things that we would like to express. I think Matt Waddington talked about similar things at the last call as well, though I might have misunderstood what he said; at any rate, the syntax proposed in that link probably is closer to the kind of 'if ... then ...' syntax that Matt had been talking about.

But of course, there might be other reasons for preferring HENCE LEST.

mengwong: Nice, thanks for re-posting this. I remember seeing it go by previously and thinking we should discuss it together as a group when the time was right. Now the time is right!

mengwong: Matt mentioned he didn't see a requirement for representing a multiplicity of jurisdictions within the same document. However the more general notion of separation – of the commitment of an offence from the penalty that it attracts – remains valid, so I would like to preserve that notion in any future modified form of your proposal.

mengwong:
obsoleted by #241, so we can close this

doc/regulative.md
mengwong: I think the functionality described in https://github.com/smucclaw/lam4/blob/88b52216a48a14bcd819aab953da59241b2dffa7/docs/proposal_normative_constructs_aug26_2024.md?plain=1#L218 is now accommodated in the latest regulative-proposal doc so let's go with that.

I forgot to mention an additional point – I wanted the coupling between THEN/ELSE and HENCE/LEST to be explicit. I was worried about a dangling ELSE in the same way Haskell insisted that every IF/THEN have an ELSE.

In next round of work I want to try implementing the offence/offence/offence/penalty pattern in the latest syntax so we can see how we feel about it.

PR #219, opened 2025-03-10
Parse and linearize NLG annotations

fendor:
### Introduce naive linearization of JL4 expressions

Linearization is the process of translating an abstract syntax tree into
a readable version of it.

Our first approximation, introduces the 'LinTree' type, which is a flat
representation of tokens produced from linearizing an 'Expr Resolved'.
We do not attempt to produce truly human-readable text, but rather try
to be predictable.

### Extend the parser to further process NLG annotations

So far, NLG annotations were treated as single tokens that we do not
look at in detail. This is insufficient, as the user can provide a
custom NLG annotation which references variables, delimited by `%`,
that are then incorporated by the NLG linearization step.

For example:

    @nlg function %a% takes parameter %`the param`%

This NLG annotation references the variable `a` and the variable
`` `the param` ``.
Since we are already parsing the names, we can also provide better
syntax highlighting for the `Name`s that occur in an NLG annotation.

Thus, we perform ad-hoc lexing and parsing of NLG annotation during the parser
step, where we replace the single NLG token with the parsed result.
The new NLG annotations store which `Name`s they are referencing,
s.t., we can later perform scope checking of the names references in the
NLG annotation.

In the typechecker, we need to perform scope checking at the right
locations to make sure all variables are actually in scope.
This is currently rather hackily done, and likely needs some more
thoughts.

An example how the syntax highlighting looks:
image


jl4-core/src/L4/TypeCheck/Types.hs
```diff
@@ -210,3 +212,114 @@ instance Alternative Check where
...
+addError :: CheckError -> Check ()
+addError e = do
+  ctx <- use #errorContext
+  with (MkCheckErrorWithContext e ctx)
+
+choose :: [Check a] -> Check a
+choose = asum
+
+anyOf :: [a] -> Check a
+anyOf = asum . fmap pure
```
MangoIV, line 242: this definition is bogus, (or at least misleadingly named and defined) I think.

MangoIV: it takes the first definition that succeeds, which must be the first element in the list because `pure x` always succeeds. I think it's equivalent to 
```haskell
\case [] -> empty; (x : _) -> pure x
```

jl4-core/src/L4/Syntax.hs
```diff
@@ -445,16 +451,44 @@ data Comment = MkComment Anno [Text]
...
+--
+-- * @Invalid@ 'MkInvalidNlg'.
+-- * @Parsed@ 'MkParsedNlg'.
+--   A valid annotation which we have processed to figure out the
+--   'Name's that occur in it.
+-- * @Resolved@ 'MkResolvedNlg'.
+--   A resolved annotation is a 'MkParsedNlg' annotation which contained only
+--   resolvable 'Name's.
+data Nlg =
+    MkInvalidNlg Anno
```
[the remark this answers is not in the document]
MangoIV: ah maybe not, probably okay to fail here :)

jl4-core/src/L4/TypeCheck/Annotation.hs
```diff
@@ -0,0 +1,239 @@
...
+
+nlgDecide :: Decide Resolved -> Check (Decide Resolved)
+nlgDecide (MkDecide ann tySig appForm body) =
+  MkDecide ann
+    <$> nlgTypeSig tySig
+    <*> nlgAppForm appForm
+    <*> nlgExpr body
+
+nlgExpr :: Expr Resolved -> Check (Expr Resolved)
+nlgExpr = \case
```
MangoIV, line 49: i think we ought to at least abstract away the binary operator cases...

jl4/examples/not-ok/tc/tests/parse-error1.nlg.golden
MangoIV, line 1: this doesn't look right.

jl4/examples/not-ok/tc/tests/parse-error2.nlg.golden
MangoIV, line 1: dito

jl4/examples/not-ok/tc/tests/parse-error3.nlg.golden
MangoIV, line 1: dito

jl4/examples/not-ok/tc/tests/parse-error4.nlg.golden
MangoIV, line 1: dito

jl4/examples/ok/tests/contract.nlg.golden
MangoIV, line 1: nice!

jl4/examples/ok/tests/nlg_lin0.nlg.golden
MangoIV, line 1: is it normal that the value of `foo` doesn't show up?

jl4/examples/ok/tests/nlg_lin2.nlg.golden
MangoIV, line 1: nice!

jl4/tests/Main.hs
```diff
@@ -94,20 +94,34 @@ jl4ExactPrintGolden dir inputFile = do
...
+  let output_ = case moutput of
+        Nothing -> "Failed to typecheck\n"
+        Just checkResult ->
+          let
+            mod' = checkResult.module'
+            directives = toListOf (gplate @(Directive Resolved)) mod'
+            dirExprs = mapMaybe (\case
+              Eval _ e -> Just e
+              Check _ e -> Just e
+              ) directives
```
MangoIV, line 110:
```suggestion
            dirExprs = map (\case
              Eval _ e -> e
              Check _ e -> e
              ) directives
```

jl4/tests/Main.hs
```diff
@@ -94,20 +94,34 @@ jl4ExactPrintGolden dir inputFile = do
...
-    Left _err -> pure "Failed to parse"
-    Right (prog, warns) -> do
-      null warns `shouldBe` isOk
-      pure $ prettyNlgOutput prog warns
+  (errs, moutput) <- oneshotL4ActionAndErrors inputFile \nfp -> do
+    let uri = normalizedFilePathToUri nfp
+    _ <- Shake.addVirtualFileFromFS nfp
+    Shake.use Rules.SuccessfulTypeCheck uri
+  let output_ = case moutput of
+        Nothing -> "Failed to typecheck\n"
```
MangoIV, line 102: this is probably where the issue comes from that the parsing failed changed to typecheck failed: you need to actually check that the parsing rule failed explicitly (this is done in this module already, so you can copy that, I guess :))

jl4/tests/Main.hs
```diff
@@ -94,20 +94,34 @@ jl4ExactPrintGolden dir inputFile = do
...
+          in
+            Text.unlines $ fmap Nlg.simpleLinearizer dirExprs
+  let output =
+        if isOk
+          then output_
+          else output_ <> "\n" <> Text.unlines (fmap Text.strip errs)
   pure
     Golden
-      { output = output_
+      { output = output
```
MangoIV, line 119:
```suggestion
      { output
```

MangoIV (changes requested):
most of my remarks are of course just nitpicks.

jl4-core/src/L4/Nlg.hs
```diff
@@ -0,0 +1,377 @@
...
+      ]
+    Or _ e1 e2 -> hcat
+      [ lin e1
+      , text "and"
+      , lin e2
+      ]
+    Implies _ e1 e2 -> hcat
+      [ lin e1
+      , text "and"
+      , lin e2
```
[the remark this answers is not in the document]
fendor: copy paste error, thanks!

jl4-core/src/L4/TypeCheck/Types.hs
fendor: @kosmikus ^

jl4/tests/Main.hs
fendor: I don't really care about the parsing any more, since NLG annotations need to be type checked as well. Thus, I improved the message, but intend to keep it like this for now.

jl4/examples/ok/tests/nlg_lin0.nlg.golden
fendor: Yes! The linearization doesn't report results right now.

jl4-core/src/L4/Parser.hs
```diff
@@ -1309,9 +1435,9 @@ epaToCluster p = CsnCluster
   , trailing = mkHiddenConcreteSyntaxNode p.trailingTokens
   }
 
-epaToHiddenCluster :: (HasField "range" t SrcRange) => Epa_ t a -> CsnCluster_ t
+epaToHiddenCluster :: (HasField "range" t SrcRange, Show t) => Epa_ t a -> CsnCluster_ t
```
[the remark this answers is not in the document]
fendor: Great catch!

jl4-core/src/L4/Nlg.hs
```diff
@@ -0,0 +1,377 @@
...
+      ]
+    Proj _ e1 e2 -> hcat
+      [ lin e1
+      , possessive "s"
+      , -- Resolved can't use 'lin', as it doesn't have an 'Anno'
+        linearize e2
+      ]
+    Var _ v -> linearize v
+    Lam _ _ _ -> mempty
```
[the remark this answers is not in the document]
fendor: Very good catch. Dunno how to linearize this, though

jl4-core/src/L4/Lexer.hs
```diff
@@ -166,9 +175,15 @@ annotations = ["nlg", "ref", "ref-map", "ref-src"]
...
-    <|> inlineAnno "[" "]"
+    <|> inlineAnno (Text.singleton nlgInlineAnnotationOpenChar) (Text.singleton nlgInlineAnnotationCloseChar)
   <?> "Natural Language Generation Annotation"
 
+nlgInlineAnnotationCloseChar :: Char
+nlgInlineAnnotationCloseChar = ']'
+
+nlgInlineAnnotationOpenChar :: Char
+nlgInlineAnnotationOpenChar = '['
+
```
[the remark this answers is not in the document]
fendor: It might be, but I feel a little better :D

jl4-core/src/L4/Nlg.hs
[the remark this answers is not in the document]
MangoIV: maybe something like "given x and y" or "let x and y"

jl4-core/src/L4/Lexer.hs
```diff
@@ -410,10 +465,14 @@ initialTokenState uri txt =
...
-mkPosTokens uri txt rtoks =
-    evalState (traverse go rtoks) (initialTokenState uri txt)
+mkPosTokens :: Maybe SourcePos -> NormalizedUri -> Text -> [RawToken] -> [PosToken]
+mkPosTokens sourcePosOffset uri txt rtoks =
+    evalState (traverse go rtoks) tokenSt
   where
+    tokenSt = case sourcePosOffset of
+      Nothing -> initialTokenState uri txt
+      Just offset -> initialTokenState uri txt
+        & #posState %~ (\p -> p { pstateSourcePos = offset} )
```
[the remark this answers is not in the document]
fendor: Indeed, looks more readable now, thanks!

MangoIV (approved):
woot woot!

PR #221, opened 2025-03-10
cli refactoring to use language server

MangoIV:


MangoIV:
I'm not entirely happy with the result of this, even thought I think it's an improvement. in future I would like the take apart the ide part and the build system part a bit more but that is significantly more work than expected. It would also be nice to emit a trace of the build system, i.e. to get all intermediate results, etc.

cabal.project
```diff
@@ -1,4 +1,6 @@
-packages: 
-  ./jl4
+packages:
+  ./jl4-core
+  ./jl4-lsp
+  ./jl4-cli
```
kosmikus, line 4: I think renaming all the example paths is bad for discoverability. Perhaps we should keep `jl4-cli` as `jl4`, then the examples would not have to be moved?

jl4-cli/app/Main.hs
```diff
@@ -0,0 +1,53 @@
...
+  oneshotL4Action (cmapWithPrio IdeLog recorder) curDir \_ ->
+    for_ options.files \fp -> do
+      let nfp = toNormalizedFilePath fp
+          uri = normalizedFilePathToUri nfp
+      Shake.addVirtualFileFromFS nfp
+      mtc <- Shake.use Rules.TypeCheck  uri
+      mep <- Shake.use Rules.ExactPrint uri
+      case (mtc, mep) of
+        (Just tcRes, Just ep) | tcRes.success -> logWith recorder Info $ ExactPrint ep
+        (_, _) -> logWith recorder Error $ CheckFailed uri
```
kosmikus, line 43: Looks nice.

jl4-cli/examples/not-ok/tests/parse-error1.ep.golden
```diff
@@ -0,0 +1,13 @@
+Info | updateFileDiagnostics published different from new diagnostics - file diagnostics:
+File:<<file>>
```
kosmikus, line 2: `<<file>>` is weird? Why would the filename not be available? If we ever have multi-module tests, this should be printed anyway.

jl4-cli/examples/not-ok/tests/parse-error1.ep.golden
```diff
@@ -0,0 +1,13 @@
...
+  Hidden:   no
+  Range:    12:1-12:1
+  Source:   parser
+  Severity: DiagnosticSeverity_Error
+  Code:     <none>
+  Message: 
+    |
+    12 | GIVEN n IS A NUMBER
+    | ^^^^^
```
kosmikus, line 11: Looks like something about the spacing / alignment of the errors is being broken. Not sure how easy this would be to fix.

jl4-cli/tests/Main.hs
```diff
@@ -3,33 +3,39 @@ module Main where
...
 
 main :: IO ()
 main = do
   dataDir <- getDataDir
   let examplesRoot = dataDir </> "examples"
   exampleFiles <- sort <$> globDir1 (compile "**/*.l4") examplesRoot
   hspec $ do
-    forM_ exampleFiles $ \ inputFile -> do
-      let testCase  = makeRelative examplesRoot inputFile
+    forM_ exampleFiles $ \inputFile -> do
```
kosmikus, line 37: Hmm. Removing my helpful spaces after lambdas, I see ...

jl4-cli/tests/Main.hs
```diff
@@ -55,11 +61,18 @@ l4Golden dir inputFile = do
...
 jl4ExactPrintGolden dir inputFile = do
-  input <- Text.readFile inputFile
-  let output_ = JL4.exactprintProgram (takeFileName inputFile) input
+  (errs, moutput) <- oneshotL4ActionAndErrors inputFile \nfp -> do
+    let uri = normalizedFilePathToUri nfp
+    Shake.addVirtualFileFromFS nfp
+    Shake.use Rules.ExactPrint uri
+
+  let regex = Text.pack "<<file>>" <$ (
+        many (RE.satisfy isSpace) *> RE.text "file://" *> some (RE.satisfy (not . isSpace)))
```
kosmikus, line 70: Ok, as remarked above, I can see why you'd want to remove / reformat the URI, but it should be easy enough to get the basename back in?

jl4-lsp/src/LSP/Core/Types/Options.hs
```diff
@@ -121,19 +121,22 @@ defaultIdeOptions = IdeOptions
     }
 
 defaultSkipProgress :: Typeable a => a -> Bool
```
kosmikus, line 123: You could perhaps add documentation what this function is even doing overall?

kosmikus (approved):
Looks great. Thanks!

jl4-cli/examples/not-ok/tests/parse-error1.ep.golden
MangoIV: it is only like that for the goldens where the full filename includes directories that have different names depending on the machine.

jl4-cli/tests/Main.hs
MangoIV: so the name of the file itself? sure.

jl4-cli/examples/not-ok/tests/parse-error1.ep.golden
kosmikus: Fair enough, although I think we dropped the paths and kept the basenames in the past. As I said, it's possible that in the not too distant future we might have tests involving multiple modules, so in which file the error is can still be relevant.

PR #226, opened 2025-03-11
Revert "jl4-web: Update visualization (with a debounce) on text document change"

MangoIV:
Reverts smucclaw/l4-ide#223


MangoIV:
@ym-han you should try whether the result of running the packaging step produces the correct outputs. it turns out this didn't, see 
image

ym-han:
> > @ym-han you should try whether the result of running the packaging step produces the correct outputs. it turns out this didn't, see image
> 
> Ah sorry, I don't really know nix -- how can I test the packaging (in the sense of, what commands do I run)? I'm assuming you mean nix packaging and not a script in the package.json of ts-apps/jl4-web. If you can give me a quick hint I'll fix it tomorrow
> 
> EDIT: I see there's a README.md, but it's a bit cryptic for beginners

I'm guessing the thing that broke the packaging was the changing it to a SPA / disabling SSR

MangoIV:
> I'm guessing the thing that broke the packaging was the changing it to a SPA / disabling SSR

me too. 

So this hasn’t got much to do with nix. The result directory has to contain an index.html that can be served by the Webserver

MangoIV:
If the output directory changed somehow, then just notify me then I’ll change that in the derivation.

ym-han:
> > I'm guessing the thing that broke the packaging was the changing it to a SPA / disabling SSR
> 
> me too.
> 
> So this hasn’t got much to do with nix. The result directory has to contain an index.html that can be served by the Webserver

Got it, I work on trying to get the build dir to contain an index.html tomorrow!

What should I run locally, though, to make sure that the derivation does work / to test it?

PR #227, opened 2025-03-11
Revert "Refactor: Remove unnecessary use of state rune for `persistSession` in jl4-web"

MangoIV:
Reverts smucclaw/l4-ide#224


MangoIV:
@ym-han I think this PR messed up the packaging (and hence deployment) of the webapp. I will deploy from this branch and then we can fix it successively :) 

I think in future it would be good to test whether changes to packaging breaks it

Issue #207, opened 2025-03-05
nixosTests for the test of the deployment


MangoIV:
it would be good to have integration tests for testing all the services together.

PR #228, opened 2025-03-12
Fix deployment issue with PR #226

ym-han:
Turns out that to produce the index.html (when set to no-ssr), I have to set `fallback` in sveltekit config to `index.html` and remove the previous fallback html page. The SvelteKit docs weren't super clear on this.

(btw I might be doing something wrong git-wise; let me know if so!)

### DIFF
--- a/ts-apps/jl4-web/src/error.html
+++ b/ts-apps/jl4-web/src/error.html
@@ -1,3 +1,0 @@
-<h1>Unknown Error</h1>
-<p>Code %sveltekit.status%</p>
-<p>%sveltekit.error.message%</p>
--- a/ts-apps/jl4-web/svelte.config.js
+++ b/ts-apps/jl4-web/svelte.config.js
@@ -13,7 +13,7 @@ const config = {
       // these options are set automatically — see below
       pages: "build",
       assets: "build",
-      fallback: "./src/error.html",
+      fallback: "index.html",
       precompress: false,
       strict: true,
     }),


MangoIV:
seems to work. thanks for fixing.

ym-han:
> seems to work. thanks for fixing.

Thanks for pointing it out!

It would be nice to have an automated test for the deployment at some point. The deployment part is something I don't currently have a good mental model for (I've mostly only deployed simple static sites or VMs for machine learning), and so the risk of my accidentally breaking something is higher here.

(I would also be down to take some time to learn nix and make that automated test if that's a task you need to distribute away -- it's something I've been meaning to learn for a while now.)

MangoIV:
@ym-han I have an issue for that (#207) 

I don't think that allocating some time to learn nix will get you very far.

This also wasn't a nix issue at all, so knowledge of it wouldn't have saved us the trouble, it was really that the build artifacts where missing. 

If you want a rough mental model: we deploy a system configuration that spins up an nginx with a virtualhost pointing at 
1. the build output for building the web app. that is served from `/` 
2. a systemd service that runs the language server websocket server that is proxied from localhost to `/lsp` 
3. a systemd service that runs the persistance service webserver
4. a systemd service that runs the webserver that answers the function calls for the LLM stuff.

ym-han:
> @ym-han I have an issue for that (#207)
> 
> I don't think that allocating some time to learn nix will get you very far.
> 
> This also wasn't a nix issue at all, so knowledge of it wouldn't have saved us the trouble, it was really that the build artifacts where missing.
> 
> If you want a rough mental model: we deploy a system configuration that spins up an nginx with a virtualhost pointing at
> 
> 1. the build output for building the web app. that is served from `/`
> 2. a systemd service that runs the language server websocket server that is proxied from localhost to `/lsp`
> 3. a systemd service that runs the persistance service webserver
> 4. a systemd service that runs the webserver that answers the function calls for the LLM stuff.

Gotcha, thanks!

And yeah, the crux of this issue wasn't really a nix thing, more a 'how websites are served' / routes thing.

MangoIV:
i think the simplest smoke test we could do is `npm run build`, then serve it with something like `darkhttpd`, curl the routes and fail if they don't return 2xx

MangoIV:
but if we do that I can also go the extra mile and do proper nixostests.

ym-han:
> i think the simplest smoke test we could do is npm run build, then serve it with something like darkhttpd, curl the routes and fail if they don't return 2xx

This sounds nice and simple. I'll also look real quick into how people using SvelteKit tend to test for this kind of thing.

I don't know how much work 'proper nixostests' would amount to, but my hunch is that we should start with the easiest-to-implement thing first. (Unless, of course, you want to assign the task of learning nixos and 'doing proper nixostests' to me, in which case, yes, i would be delighted to have the privilege of spending some time learning it properly :)

MangoIV:
let's coordinate in the meeting tomorrow - then we can also ask Meng and Andres.

PR #231, opened 2025-03-12
make jl4-web editor pane / visualisation panel resizable

### LINKED ISSUE #212: the visualization panel cannot be resized atm

johsi-k:
Resolves #212 

* Add tailwind to jl4-web (seems fine to add it now given that we are more confident in the css infra now; also tailwind is useful for layouting things, and is required for shadcn-svelte)
* Add Shadcn-Svelte dep / infra
* Use Resizable component from Shadcn-Svelte to make editor and visualization panes resizable, with visible Handle
* Remove `postcss` dep from jl4-web -- jl4-web doesn't need this
* Refactor: Remove CSS that's not actually being applied (the `visualization-container` class)
* Style tweak: Add some bottom padding to ladder flow div container in jl4 webview panel

image


ym-han:
Just to tackle the first point for now:

> I think we previously explicitly did not pull in tailwind into this app because it was considered unnecessary, I don't quite understand why adding it is now required - afaict none of the styling actually changed

Yeah, you're right that this wasn't explained well. 

At that point in time, I was inclined to keep the CSS infra as simple as possible (in part because there was still some uncertainty re things like whether the css setup in the decision-logic-visualizer lib was working ok, as well as over things like whether there would be complications re using Tailwind as a consumer of a lib that also uses Tailwind). Our UI needs for jl4-web were also a lot simpler.

I am less uncertain now about the css infra (this is not to say there aren't kinks left to iron out; it's just to say that it's clearer to me now that the most important parts, e.g. the exporting of the css from the dlv library, work, and how they work), and so am not as anxious now about keeping things ultra-minimalistic. And the UI needs for jl4-web have grown (e.g., we now need a resizer component) and, as far as I can see, will likely continue to grow (e.g., we might want to have more sub-panels in the right pane). 

This, of course, still doesn't *require* that we use Tailwind: we could, for instance, use the underlying thing that the shadcn-svelte library uses for their resizer, instead of shadcn-svelte's resizer. It's just that, based on my admittedly limited experience with shadcn-svelte, it seems like a good library, and there's enough productivity benefits for me from using Tailwind, for the kinds of UI that I'm guessing we want to use for jl4-web, that I'm thinking the benefits of using it now outweigh the cost of additional complexity.

I'm not wedded to this, though, and Jo Hsi is not a diehard fan of Tailwind either. So we can definitely refactor this to not use Tailwind.

MangoIV:
sounds fine, yes, thank you for the reasoning!

johsi-k:
I've removed the handle for now. Initially, we chose the visible handle to improve affordance for users unfamiliar with a split editor environment, but I think the bare divider feels more idiomatic.

ym-han:
@MangoIV it's fine to merge this in now, yeah?

ym-han:
Another example of an editor with multiple resizable panes, for future reference (not saying we have to do anything with this now):
https://haskell-playground.co.uk/editor

image

ym-han:
> @ym-han yeah it's fine by me. could we still make the bar a few pixels wider?

Jo Hsi will look into making the bar a few pixels wider.

PR #237, opened 2025-03-13
Experiment with a wider divider between editor and viz panes

johsi-k:
Screenshot 2025-03-13 at 23 47 25


ym-han:
Seems fine to me. I'm not sure that an inline style is the best way to do this, but it also doesn't matter (because it won't affect anything else -- we can look into this more if we end up needing to add a lot more styles to the resizable handle or something).

ym-han (approved):
Seems fine to me. I'm not sure that an inline style is the best way to do this, but it also doesn't matter (because it won't affect anything else -- we can look into this more if we end up needing to add a lot more styles to the resizable handle or something).

johsi-k:
@ym-han we could expose width as an explicit prop, but since we don't anticipate needing to customize it frequently, I opted for inline styles instead. We can revisit best practices if we need to add more styles in the future.

@MangoIV thank you!

PR #239, opened 2025-03-14
Viz, Monaco, Webview: Make the updating of the FunDeclLirNode in response to cmd execution simpler and more robust. Do more to clean up resources and facilitate garbage collection.

ym-han:
Problem: the monaco jl4-web can sometimes get sluggish and hog a ton of memory, especially when using the viz. This was making it hard to work on jl4-web.

This PR tries to address that by doing more to clean up resources and facilitate garage collection (there was a lot of low hanging fruit here, since I had made a conscious effort to avoid prematurely working on this kind of optimization). This should be especially helpful when it comes to the visualizer, because I had used a lot of immutable data and done a lot of copying instead of mutating in place.

Along the way, I also made the updating of the FunDeclLirNode in response to cmd execution  in `jl4-web` and `webview` simpler and more robust. That part of the codebase should hopefully also be easier for other people to understand now.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14326,6 +14326,7 @@
         "monaco-editor": "npm:@codingame/monaco-vscode-editor-api@^14.0.4",
         "monaco-editor-wrapper": "^6.4.0",
         "monaco-languageclient": "^9.4.0",
+        "runed": "^0.23.3",
         "vscode": "npm:@codingame/monaco-vscode-extension-api@^14.0.4",
         "vscode-languageclient": "^9.0.1",
         "vscode-ws-jsonrpc": "^3.4.0"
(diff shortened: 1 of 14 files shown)


MangoIV (approved):
Looks good. Can you explain why the GC doesn't do this automatically perhaps?

ym-han:
> Looks good. Can you explain why the GC doesn't do this automatically perhaps?

Done, thank you!

Also, just as a quick fyi:
* I added the monaco editor dispose bit because I saw some examples on github that did that, but I'm not sure it's actually necessary. (But it also didn't seem like it'd hurt to add it.) I'll look into it more in the future.
* I've checked that this seems to be making a difference re whether certain LirNodes are GC'd, but there may be 1 - 2 lines of code that might not be necessary for this. Also, there's still more that can be done here -- this is just the first attempt to clear some low-hanging fruit.

PR #241, opened 2025-03-16
instead of hardcoding two examples for LLM reasoner API, support YAML facility for loading from disk.

mengwong:
this also adds the parameter names to the "required" property.

this PR also includes the documentation-oriented #209 as that branch was needed for the `doc/tutorial-code/fruit.yaml`


kosmikus:
1. Could we split this so that the documentation PR and the decision-service PR are separate? They seem entirely independent, and we should review them separately, making it easier to merge one part when it's ready.
2. Tutorial code seems untested. It would be good if we could make sure this is being regression-tested, otherwise these will be files that can easily bitrot.
3. Tests currently fail, probably because you changed the API of the decision service without adapting the tests?

mengwong:
ok, i will work on these points of feedback

> 
> Could we split this so that the documentation PR and the decision-service PR are separate? They seem entirely independent, and we should review them separately, making it easier to merge one part when it's ready.
> Tutorial code seems untested. It would be good if we could make sure this is being regression-tested, otherwise these will be files that can easily bitrot.
> Tests currently fail, probably because you changed the API of the decision service without adapting the tests?
>

mengwong:
obsoleted by #258 and #259, closing.

PR #243, opened 2025-03-17
Improve type synonym expansion.

### LINKED ISSUE #233: Type synonym expansion seems not to be working correctly
I need to create a minimal example for this, but
```
DECLARE Pred a IS A FUNCTION FROM a TO BOOLEAN
```
seems to cause problems if functions that are assumed to be of `Pred` type are going to be applied.

kosmikus:
Fixes #233.

The problem was that type synonym expansion is being done in unify, but for the purposes of generating error messages, we check function applications using a specialised function, in which we did not perform type synonym expansion.


MangoIV (approved):
looks plausible.

PR #245, opened 2025-03-17
Viz: Improve code quality of some of the svelteflow-related code (docs, names, and future-proof the getting of the sf id)

ym-han:
* Rename `types.svelte.ts` to the clearer `svelteflow-types.ts`
* Clean up and improve docstrings for SvelteFlow graph related things. 
In particular, explicitly document some of the key assumptions / invariants in the rendering of the SF graph from the intermediate LadderLir representation wrt state changes.
* Add section headers to some of the svelte flow related code and standardize / clean up some of the code patterns
* Make the getting of the SF id from an SF node less fragile / more future-proof, by using a function (and thereby doing some info hiding), instead of the specific `id` property on the object.

The first three things should help a lot with making the codebase clearer to other people and future versions of myself.


MangoIV (approved):
i think this is okay

PR #251, opened 2025-03-18
Viz: Fix and improve Tailwind config + theme css in `decision-logic-visualizer` package (has to do with a TW v3 vs v4 difference). Add and adapt css vars and styles needed for shadcn svelte

ym-han:


MangoIV (approved):
lgtm

PR #253, opened 2025-03-18
IMPORT statements

### LINKED ISSUE #187: Allow imports of other files
Add a way to import other files.

Let's not tie this up with namespacing just yet. All (top-level) declarations from imported files should become available.

MangoIV:
Resolves #187 
- [x] evaluation (merging of environments is missing)
- [x] breaking cyclic dependencies
- [x] sanitize tests


MangoIV:
This is not ready to merge, there's a problem with `fib.l4`

jl4-lsp/src/LSP/Core/IdeConfiguration.hs
```diff
@@ -86,12 +86,17 @@ isWorkspaceFile uri =
...
+
           let toText = getUri . fromNormalizedUri
-          return $
-            any
-              (\root -> toText root `isPrefixOf` toText (normalizedFilePathToUri file))
-              workspaceFolders
+          pure case mconfig of
+            -- NOTE: if we are not in an editor context, we assume that every file is part
+            -- of our "workspace" as in this case there's not really a concept of a workspace
+            Left (_ :: SomeException) -> True
```
fendor, line 95: I assume we can be more accurate than `SomeException`? I think `safe-exceptions` behaves correctly, but I do feel much better when we are explicitly catching the exception that might occur.

jl4-lsp/src/LSP/Core/IdeConfiguration.hs
```diff
@@ -86,12 +86,17 @@ isWorkspaceFile uri =
...
+            Right config ->
+              any
+                (\root -> toText root `isPrefixOf` toText (normalizedFilePathToUri file))
```
fendor, line 98: I know you didn't write this, but I am pretty sure this check is wrong.

fendor (approved):
Approving so I don't block this forever, but I would like to investigate the cycle detection further in the future.

jl4-core/src/L4/Evaluate.hs
```diff
@@ -82,7 +106,7 @@ buildEvalTrace = go []
 
 data EvalException =
     RuntimeScopeError Resolved -- internal
-  | RuntimeTypeError -- internal
+  | RuntimeTypeError String -- internal
```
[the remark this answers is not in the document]
MangoIV: that would require some more refactoring, maybe let's keep this as futurework?

jl4-core/src/L4/TypeCheck.hs
```diff
@@ -1792,10 +1811,11 @@ prettyOptionallyNamedType (MkOptionallyNamedType _ (Just r) t) =
 --
 -- TODO: eventually, we will have to print a file path here for potentially external locations
 prettyResolvedWithRange :: Resolved -> Text
-prettyResolvedWithRange r =
+prettyResolvedWithRange r = do
+  let u = getUnique r
   case rangeOf (getOriginal r) of
     Nothing    -> prettyLayout r <> " (predefined)"
-    Just range -> prettyLayout r <> " (defined at " <> prettySrcRange Nothing (Just range) <> ")"
+    Just range -> prettyLayout r <> " (defined at " <> (fromNormalizedUri u.moduleUri).getUri <> ":" <> prettySrcRange Nothing (Just range) <>  ")"
```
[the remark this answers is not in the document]
MangoIV: there's many variations of this. sometimes you need the text, sometimes you need the string, sometimes you need the Uri, sometimes you need the NormalizedUri... I think what makes it awkward is `-XNoFieldSelectors`

jl4-lsp/src/LSP/Core/IdeConfiguration.hs
MangoIV: I also have a weird feeling about it but I don't immediately see what's wrong about it (uris should be normalized, they should both be file uris)

jl4-core/src/L4/Evaluate.hs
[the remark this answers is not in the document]
kosmikus: Yeah, run-time type errors should be "internal errors", so I don't think they need to be extremely beautiful.

jl4-core/src/L4/Evaluate.hs
```diff
@@ -12,26 +21,39 @@ import L4.Utils.RevList
 
 import Data.Either
 
-newtype Eval a = MkEval (EvalState -> (Either EvalException a, EvalState))
-  deriving (Functor, Applicative, Monad, MonadError EvalException, MonadState EvalState)
-    via ExceptT EvalException (StateT EvalState Identity)
+newtype Eval a = MkEval (EvalState -> EvalEnv -> (Either EvalException a, EvalState))
```
kosmikus, line 24: Still not extremely happy that we're adding a `Reader` component for just the URI, but ok ...

kosmikus (approved):

jl4-lsp/src/LSP/Core/IdeConfiguration.hs
fendor: The `toText` strips the trailing slash, so this check:

```haskell
> toText = getUri . fromNormalizedUri
> toNUri = toNormalizedUri . filePathToUri
> toText (toNUri "/home/hugin/Documents/haskell/cclaw/l4-ide/jl4-lsp/src/LSP/Core/") `Text.isPrefixOf` toText (toNUri "/home/hugin/Documents/haskell/cclaw/l4-ide/jl4-lsp/src/LSP/Core-Not-Dir.hs")
True
```
This should be `False` but returns `True`.

PR #255, opened 2025-03-20
Fix #211: Added css to move command palette to prevent cut off

### LINKED ISSUE #211: the command panel opens in the middle of the screen and is cut off

Meowyam:


ts-apps/jl4-web/src/style.css
```diff
@@ -61,3 +61,16 @@
     border: 1px solid grey;
   }
 }
+
+.quick-input-widget {
```
ym-han, line 65: These styles seem quite specific to the monaco editor; for that reason, it'd be better to put them in the `<style> ... </style>` block in `jl4-web/src/routes/+page.svelte`.

ts-apps/jl4-web/src/style.css
```diff
@@ -61,3 +61,16 @@
...
+  position: absolute !important;
+  left: 25px !important;
+  right: auto !important;
+  top: 25px !important;
+}
+
+div[widgetid="editor.contrib.quickInputWidget"] {
```
ym-han, line 72: It's worth adding a quick comment to the effect that `editro...quickInputWidget` is an id internal to Monaco, if that is the case, as opposed to a public API.

ym-han (changes requested):
I think it's better to put the styles in the `<style> ... </style> block in jl4-web/src/routes/+page.svelte`.

kosmikus:
Apparently, there are CI errors regarding code style. Could you please fix these?

PR #258, opened 2025-03-21
[docs] added language tutorial and proposal for regulative rules

mengwong:
Extracted from #241

Next steps: move tutorial-code into the jl4/examples dir, with tests.

> Tutorial code seems untested. It would be good if we could make sure this is being regression-tested, otherwise these will be files that can easily bitrot.


mengwong:
now that tests pass, i'm going to go ahead and merge this unless there are any objections

PR #259, opened 2025-03-21
decision service tweaks: support --sourcePaths to load yaml from disk; finesse `Inputs` management; add support for CORS.

mengwong:
Extracted from #241 

TODO:
> Tests currently fail, probably because you changed the API of the decision service without adapting the tests?

### DIFF
--- a/jl4-decision-service/README.md
+++ b/jl4-decision-service/README.md
@@ -13,9 +13,21 @@ In a future iteration the backend will
 
 Run the decision service backend locally:
 
-    cabal run jl4-decision-service-exe -- --port 8081 --serverName http://localhost:8081/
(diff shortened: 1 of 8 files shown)


mengwong:
tests are failing due to the JSON producing `required` fields. However the schema definition does not know about those `required` fields. I'm afraid I am not familiar enough with the internals of Data.OpenApi and Servant / Swagger etc to be able to modify the schema 

image

The fix needs to happen somewhere around https://github.com/smucclaw/l4-ide/blob/mengwong/decisionservice/jl4-decision-service/src/Schema.hs#L128

mengwong:
now that tests pass, i'm going to go ahead and merge this unless there are any objections

PR #267, opened 2025-03-24
[fix] do prefix check that is filepath aware

MangoIV:


kosmikus:
LGTM, but I'd prefer if @fendor has another look.

fendor (approved):

PR #273, opened 2025-03-24
Fix type checker

### LINKED ISSUE #266: Definition of reverse makes jl4 loop
```
GIVEN
  a    IS A TYPE
  list IS A LIST OF a
GIVETH A LIST OF a
reverse list MEANS go EMPTY list
  WHERE
    go acc l MEANS
      CONSIDER l
      WHEN EMPTY            THEN acc
      WHEN x FOLLOWED BY xs THEN go (x FOLLOWED BY acc) xs
```
Causes a loop in the compiler pipeline, I am not sure where.

kosmikus:
Fixes #266.


kosmikus:
I can try to explain this to you in more detail, yes.

PR #283, opened 2025-03-26
configure cors middleware to handle OPTIONS correctly and also content-type

mengwong:


MangoIV:
if we really need to answer OPTIONS requests, then I doubt this solves the issue. There's no route for OPTIONS requests atm. I don't think the middleware adds it.

MangoIV (changes requested):
I have tested locally and it doesn't solve the issue, `OPTIONS` is not part of the API.

MangoIV:
I think it would be good to look at the error that we're currently getting and then check out what we need to provide. If we really need to add the `OPTIONS` route, then we would have to add it to the servant API. :)

MangoIV:
Hey @mengwong should we merge this or should I have a look at what we need to do get the `OPTIONS` into the API? Otherwise we can talk about this tomorrow?

mengwong:
The motivation for all this –

mengwong:
on this branch when i try the execution through the swagger UI I get a different error which is more expected.

image

mengwong:
hm, now the query works, but the reasoning payload seems kind of weird, lol

image

PR #289, opened 2025-03-27
multi file support for find references

MangoIV:


jl4-core/src/L4/Evaluate.hs
```diff
@@ -127,10 +127,7 @@ step = do
   pure current
 
 newUnique :: Eval Unique
-newUnique = do
-  i <- step
-  u <- asks (.moduleUri)
-  pure (MkUnique 'e' i u)
+newUnique = MkUnique 'e' <$> step <*> asks (.moduleUri)
```
kosmikus, line 130: I don't understand why we have to make such completely random changes.

jl4-core/src/L4/Lexer.hs
```diff
@@ -42,8 +42,6 @@ data PosToken =
   deriving stock (Eq, Ord, Show, Generic)
   deriving anyclass (ToExpr, NFData)
 
-instance ToExpr NormalizedUri
```
kosmikus, line 45: Why?

jl4-core/src/L4/Syntax.hs
```diff
@@ -23,7 +23,7 @@ data RawName =
   deriving stock (GHC.Generic, Eq, Ord, Show)
   deriving anyclass (SOP.Generic, ToExpr, NFData)
 
-data Unique = MkUnique {sort :: !Char, unique :: !Int, moduleUri :: !NormalizedUri}
+data Unique = MkUnique {sort :: !Char, unique :: !Int, moduleUri :: NormalizedUri}
```
kosmikus, line 26: Why?

jl4-core/src/L4/TypeCheck/Types.hs
```diff
@@ -225,10 +226,7 @@ step = do
   pure current
 
 newUnique :: Check Unique
-newUnique = do
-  i <- step
-  u <- asks (.moduleUri)
-  pure (MkUnique 'c' i u)
+newUnique = MkUnique 'c' <$> step <*> asks (.moduleUri)
```
kosmikus, line 229: Why?

jl4/examples/legal/tests/britishcitizen2.golden
```diff
@@ -1,7 +1,7 @@
 Parsing successful
 Typechecking successful
 Evaluation successful
-britishcitizen2.l4:54:7-17:
+ britishcitizen2.l4:54:7-17:
```
kosmikus, line 4: This looks wrong. We should check where this space is coming from, but it certainly shouldn't be there and needlessly change all the test results.

kosmikus (changes requested):
I think the test output change should definitely be changed. The rest is basically just me either not seeing the need of certain rewrites or being genuinely puzzled why they're needed.

MangoIV:
@kosmikus the changes have been caused by my back and forth, I should’ve checked the final diff!

jl4/examples/legal/tests/britishcitizen2.golden
MangoIV: Yeah that was lazy, you’re totally right

jl4-core/src/L4/Lexer.hs
MangoIV: Move it further down the dependency tree. We dont particularly need it (again, because I reverted some changes) but it also can’t hurt to have it as far down as possible for the future.

jl4-core/src/L4/Syntax.hs
MangoIV: Accident

PR #298, opened 2025-03-28
VSCode Webview: Remove unnecessary dep from `webview` package

ym-han:


fendor:
Don't we need to regenerate the package-lock.json file after?

ym-han:
It's a priori surprising to me that the CI didn't complain. I'll look more into this when I have time.

PR #302, opened 2025-03-28
Allow `NOT OF` as well as `NOT`.

kosmikus:
This makes it more consistent with other prefix functions. We keep `NOT` as a keyword / prefix operator rather than as an ordinary function call though, to e.g. keep `NOT NOT x` working out of the box, which would otherwise be interpreted as a call to `NOT` supplying two arguments.


MangoIV (approved):
LGTM

PR #303, opened 2025-03-28
[feat] report error on stuck on term instead of throwing internal errors

MangoIV:
I have yet to reproduce the runtime scope errors. Perhaps they're already ruled out by these fixes?


jl4/tests/Main.hs
```diff
@@ -158,7 +158,7 @@ checkFile isOk file = do
 
  where
   typeErrorToMessage err = (JL4.rangeOf err, JL4.prettyCheckErrorWithContext err)
-  evalDirectiveResultToMessage (JL4.MkEvalDirectiveResult r res _) = (Just r, [either Text.show Print.prettyLayout res])
+  evalDirectiveResultToMessage (JL4.MkEvalDirectiveResult r res _) = (Just r, [either JL4.prettyEvalException Print.prettyLayout res])
```
kosmikus, line 161: As said above, ideally reuse `cliErrorMessage` here.

kosmikus (approved):
Apart from error message formatting, looks good to me. Let's get this merged. We can open another issue if there are situations left where evaluation on `Assume` causes strange errors.

PR #305, opened 2025-03-28
TS CI: Check if the package.json is in sync with lockfile. (And remove a dependency from ladder/dlv package.)

ym-han:
The check seems to fail, as desired, in at least one case when package.json is not in sync with the lockfile:

https://github.com/smucclaw/l4-ide/actions/runs/14133585682/job/39599821973?pr=305


MangoIV (approved):
Very good, thanks

PR #310, opened 2025-03-30
Fixing some bugs

MangoIV:
- [x] make `MAYBE AKA OPTIONAL`  complete.

### DIFF
--- a/jl4-core/src/L4/TypeCheck.hs
+++ b/jl4-core/src/L4/TypeCheck.hs
@@ -76,7 +76,6 @@ module L4.TypeCheck
 
 import Base
 import qualified Base.Map as Map
-import qualified Base.Set as Set
 import qualified Base.Text as Text
 import L4.Annotation
 import L4.Parser.SrcSpan (prettySrcRange, prettySrcRangeM)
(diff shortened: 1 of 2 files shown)


jl4-core/src/L4/TypeCheck.hs
```diff
@@ -143,11 +142,14 @@ doCheckProgramWithDependencies checkState moduleUri program =
...
 combineEnvironmentEntityInfo :: Environment -> EntityInfo -> [(Name, CheckEntity)]
 combineEnvironmentEntityInfo env ei =
-  let
-    uniques :: Set Unique
-    uniques = Set.fromList (concat (snd <$> Map.toList env))
-  in
-    snd <$> filter (flip Set.member uniques . fst) (Map.toList ei)
+    foldMap (uncurry lookupUnqs) $ Map.toList env
+    where
+      lookupUnqs rn unqs = foldMap (maybeToList . fmap (first (replaceRawName rn)) . (ei Map.!?)) unqs
```
fendor, line 147:
```suggestion
      lookupUniques rn uniques = foldMap (maybeToList . fmap (first (replaceRawName rn)) . (ei Map.!?)) uniques
```

fendor: I don't think the abbrevation is worth it :)

jl4-core/src/L4/TypeCheck.hs
```diff
@@ -143,11 +142,14 @@ doCheckProgramWithDependencies checkState moduleUri program =
...
+
+      -- NOTE: the reason why we do this is because the CheckEntity doesn't contain the original name
+      -- e.g. if you have `foo AKA bar`, the CheckEntity will always contain `bar`. However, the environment
+        -- still has the correct name.
```
fendor, line 151:
```suggestion
      -- still has the correct name.
```

jl4-core/src/L4/TypeCheck.hs
fendor: Also, might be worth splitting over multiple lines.

fendor: Type signature might be helpful as well.
This line of code feels overcomplicated to me.

fendor (approved):

PR #316, opened 2025-03-31
Expose NLG information in Hover text

fendor:
We expose the NLG linearisation to the `InfoTree`. This makes it easy to discover and can be used as a form of documentation.

Additionally, add a bunch of type and name resolution to the AST after type checking.

Similarly to the NLG annotations, we need to be careful to add the additional type information when the respective type information is actually available.
For example, when new variables are brought into scope, for example in:

```
  ...
  CONSIDER x
  WHEN y FOLLOWED BY ys THEN y
  OTHERWISE 0
```

then the `Resolved` `y` is only in scope within the `inferBranch` code, i.e. `THEN y` part. Thus, we need to make sure to add type information to `y` after we have typechecked the body of branch.

The same logic is applied to anything that brings a new binder into scope.
As a rule of thumb, whenever we have a `scope` within the typechecker, some extra logic will need to be added.

### DIFF
--- a/jl4-core/src/L4/HoverInfo.hs
+++ b/jl4-core/src/L4/HoverInfo.hs
@@ -3,13 +3,14 @@ module L4.HoverInfo where
 
 import Base
 import L4.Annotation
-import L4.Lexer (PosToken(..))
-import L4.Parser.SrcSpan (SrcPos(..), SrcRange(..), inRange)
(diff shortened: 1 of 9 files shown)


MangoIV (approved):
I think this is okay. I think the manual traversal bits are getting a bit out of hand but we talked about that already :)

PR #318, opened 2025-04-01
Toolbar

### LINKED ISSUE #254: Viz: Make a simple menu / toolbar, with some way to switch between various modes
In the short term I think the following kind of menu should be good enough.

`Drop down | zen mode switch`

where | is a visual separator

The drop down should have a ‘Mode’ label, with 'Select Paths' and '#WhatIf' as the options in the short term

I recommend using shadcn svelte if possible. I use it in #252  --- you can check that out to see where I’ve put the shadcn svelte ui primitives.

No need to worry about wiring it up to anything — just focus on the layouting and css and I can take care of the rest.

Meowyam:
A photoshop-style toolbar that has three buttons: Select paths, Whatif and Zen mode. It's draggable. Icons can be changed. Closes #254


ym-han:
Further sub-tasks:

- [ ] @johsi-k to look into whether `z-[9999]` really is a good idea -- feels like a hack to me
- [ ] Instead of using toolbar in the route page, probably use it in flow-base.svelte (@johsi-k could try doing this)
- [ ] I will check where Toolbar.svelte should be put

ym-han:
Deprecated for now because we are using a simpler UI for the node selection, and we don't really have anything else that really really requires this.
Also, the associated PR will require some clean up work.

PR #331, opened 2025-04-03
Allow C-style comments.

### LINKED ISSUE #137: feature request: add support for C-style and Javascript-style comments
Currently our comment syntax is
```
-- two dashes
```

In future, it would be nice to also support
```
// two slashes
/* c-style */
# hash prefix
```

Why? Because the inevitabilty of AI-generated code means Copilot is going to write code with those comments anyway and it will be difficult to resist the juggernaut.

kosmikus:
Fixes #137.

### DIFF
--- a/jl4-core/src/L4/Lexer.hs
+++ b/jl4-core/src/L4/Lexer.hs
@@ -226,11 +226,12 @@ whitespace =
 
 lineComment :: Lexer Text
 lineComment =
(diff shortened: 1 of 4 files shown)


MangoIV (approved):
I think this is good. I’d probably refactor into smth similar as the nlg/ref parsers (calling the same code with different tokens) but I think this is good for only a single new set of tokens.

PR #339, opened 2025-04-07
experiments for the british citizen use case, also dev prerequisites

mengwong:


shell.nix
MangoIV, line 1: there's already a `shell.nix` - try `nix-shell nix/shell.nix`

mengwong: Thanks, will use that.

Just for the record, the compile error I got was:

```
mengwong@nixos [11:55:54] [~/src/smucclaw/mattwaddington] [mengwong/moredocs *]
-> % cabal build all
Resolving dependencies...
Error: [Cabal-7107]
Could not resolve dependencies:
[__0] trying: jl4-0.1 (user goal)
[__1] trying: base-4.18.2.1/installed-4.18.2.1 (dependency of jl4)
[__2] trying: jl4-decision-service-0.1.0.0 (user goal)
[__3] trying: servant-swagger-ui-0.3.5.5.0.1 (dependency of jl4-decision-service)
[__4] trying: file-embed-lzma-0.1 (dependency of servant-swagger-ui)
[__5] trying: lzma-0.0.1.1 (dependency of file-embed-lzma)
[__6] rejecting: lzma:+pkgconfig (conflict: pkg-config package liblzma>=5.2.2, not found in the pkg-config database)
[__6] rejecting: lzma:-pkgconfig (manual flag can only be changed explicitly)
[__6] fail (backjumping, conflict set: lzma, lzma:pkgconfig)
After searching the rest of the dependency tree exhaustively, these were the goals I've had most trouble fulfilling: base, lzma, servant-swagger-ui, file-embed-lzma, lzma:pkgconfig, jl4-decision-service, jl4
Try running with --minimize-conflict-set to improve the error message.
```

So that's why I put xz.dev in the shell.nix. I see it is already there in the nix/shell.nix. Thanks.

PR #340, opened 2025-04-07
Viz: Distinguish between compatible and incompatible paths

ym-han:


ts-apps/decision-logic-visualizer/src/lib/data/viz-expr-to-lir.ts
```diff
@@ -114,10 +120,10 @@ function transform(
       const boolvar = new BoolVarLirNode(nodeInfo, originalVar)
       const graph = vertex(boolvar.getId())
       const newEnv = new Map(env).set(originalVar.id, graph)
-      return { graph, vizExprToLir: newEnv }
+      return { graph, vizExprToLirGraph: newEnv }
```
ym-han, line 123: This following basically fixes a bug from the previous PR in the threading of the vizExprToLirGraph

ts-apps/decision-logic-visualizer/src/lib/layout-ir/ladder-graph/edge-attributes.ts
```diff
@@ -0,0 +1,195 @@
+import type { Eq } from '$lib/utils.js'
+import 'core-js/actual/set'
+
+/********************************
+      Attributes
+*********************************/
+// This isn't in principle limited to edges;
+// it's just that we will prob use LirNode methods for node data
+
```
ym-han, line 9: The main thing that's new here is the EdgeStylesContainer stuff. The diff isn't the clearest because i moved the file

ts-apps/decision-logic-visualizer/src/lib/layout-ir/paths-list.ts
```diff
@@ -0,0 +1,90 @@
+/* eslint-disable @typescript-eslint/no-unused-vars */
+import {
```
ym-han, line 2: This is basically factored out from the old ladder-lir.svelte, as part of an effort to break up the old and too-big ladder-lir.svelte file into smaller pieces

MangoIV (approved):

PR #341, opened 2025-04-07
Refactor TypeCheck phase to use explicit Scopes

fendor:
Instead of relying on mutations in the 'TypeCheck' phase, we rather want to explicitly keep track of which Identifiers are in actually in scope.

From a technical point of view, we don't keep track of 'Environment' and 'CheckEntity' in a State monad, which requires us to explicitly use scoping functions, such as 'scope' and keep track of exposed Identifiers, such as function names.

This is less elegant and makes the scopes much harder track. Instead, we keep the 'Environment' and 'CheckEntity' in a 'Reader'. The Reader has a natural scoping mechanism 'local' and delimits the scope of Identifiers more naturally.

We still need to explicitly keep track of Identifiers that need to be exported to the outside scope, but we do so now in a consistent way, instead of shotgun changes.

Most of the changes are rather mechanical, and not exciting.


jl4-core/src/L4/TypeCheck/Types.hs
```diff
@@ -162,11 +167,11 @@ runCheck' e s (MkCheck f) = f e s
...
-    { program      :: Module  Resolved
-    , errors       :: [CheckErrorWithContext]
-    , substitution :: Substitution
-    , environment  :: Environment
-    , entityInfo   :: EntityInfo
+    { program      :: !(Module  Resolved)
+    , errors       :: ![CheckErrorWithContext]
+    , substitution :: !Substitution
+    , environment  :: !Environment
+    , entityInfo   :: !EntityInfo
```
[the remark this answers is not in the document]
MangoIV: i don't really care.

jl4-core/src/L4/TypeCheck.hs
```diff
@@ -936,74 +1003,35 @@ maybeGivethType :: Maybe (GivethSig Resolved) -> Check (Type' Resolved)
...
-  traverse_ (flip makeKnown ce) rs
-
 checkExpr :: ExpectationContext -> Expr Name -> Type' Resolved -> Check (Expr Resolved)
-checkExpr ec (IfThenElse ann e1 e2 e3) t = softprune $ scope $ do
+checkExpr ec (IfThenElse ann e1 e2 e3) t = softprune $ do
   re <- checkIfThenElse ec ann e1 e2 e3 t
   let re' = setAnnResolvedType t re
   pure re'
-checkExpr ec (Consider ann e branches) t = softprune $ scope $ do
+checkExpr ec (Consider ann e branches) t = softprune $ do
```
MangoIV, line 1018: why did this `scope` disappear without replacement?

MangoIV: I guess this was a bug before?

jl4-core/src/L4/TypeCheck.hs
```diff
@@ -936,74 +1003,35 @@ maybeGivethType :: Maybe (GivethSig Resolved) -> Check (Type' Resolved)
...
   pure re'
 -- checkExpr (ParenExpr ann e) t = do
 --   re <- checkExpr e t
 --   pure (ParenExpr ann re)
-checkExpr ec (Where ann e ds) t = softprune $ scope $ do
-  rds <- traverse (nlgLocalDecl <=< inferLocalDecl) ds
-  re <- checkExpr ec e t
+checkExpr ec (Where ann e ds) t = softprune $ do
+  (rds, extends) <- unzip <$> sequentialTraverse (firstM nlgLocalDecl <=< inferLocalDecl) ds
+  re <- extendKnownMany (concat extends) do
```
MangoIV, line 1027: shouldn't `last` extends be good enough?

MangoIV (approved):
looks good!

jl4-core/src/L4/TypeCheck.hs
fendor: All scopes are explicit now, there is no one `scope` function any more. `scope` has been replaced by `errorContext` and `extendKnown*`, which modify the `ErrorContext` and `Environment`/`EntityInfo` for the given scope.

jl4-core/src/L4/TypeCheck.hs
fendor: I should document this:
`sequentialTraverse` returns a `(a, [CheckInfo])` where each `[CheckInfo]` is only what `inferLocalDecl` actually returned, not the accumulation of all intermediate scopes.

jl4-core/src/L4/TypeCheck.hs
MangoIV: Yeah I mean that I understand but I still wonder what the scope that was there before was supposed to scope over…

jl4-core/src/L4/TypeCheck.hs
MangoIV: Ah I understand. I thought that the traverse thing actually keeps all intermediate results.

MangoIV:
@fendor are we ready to merge this? Would be nice since that would make some of my tasks easier.

jl4-core/src/L4/TypeCheck.hs
fendor: It was scoping the definitions of `Where` and `Lambda`, which both brought new binders into scope. This is now managed explicitly.

kosmikus (commented):
Sorry for the late review. It looks great overall.

jl4-core/src/L4/TypeCheck/Types.hs
```diff
@@ -162,11 +167,11 @@ runCheck' e s (MkCheck f) = f e s
 
 data CheckResult =
   MkCheckResult
-    { program      :: Module  Resolved
-    , errors       :: [CheckErrorWithContext]
-    , substitution :: Substitution
-    , environment  :: Environment
-    , entityInfo   :: EntityInfo
+    { program      :: !(Module  Resolved)
+    , errors       :: ![CheckErrorWithContext]
```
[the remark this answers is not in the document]
fendor: I know it is useless, but it turned some warnings into errors for record initialisation.

PR #349, opened 2025-04-09
(recursive) qualification of sections and a refactor for the AST to make sections nested

MangoIV:
- refactoring to make sections truly nested (instead of containing their nesting level as a field) 
- qualification over (nested) sections as long as the section is already fully defined (recursive nested sections don't work yet)


MangoIV:
Don't fear - the diff is so large because of the goldens! :D

jl4-core/src/L4/Print.hs
```diff
@@ -171,33 +173,36 @@ instance LayoutPrinter a => LayoutPrinter (Import a) where
...
 instance LayoutPrinter a => LayoutPrinter (Module  a) where
   printWithLayout = \case
-    MkModule _ _ sects ->
-      vcat (fmap printWithLayout sects)
+    MkModule _ _ sect -> printWithLayout (0 :: Int, sect)
 
 instance LayoutPrinter a => LayoutPrinter (TopDecl a) where
+  printWithLayout t = printWithLayout (1 :: Int, t)
+
+instance LayoutPrinter a => LayoutPrinter (Int, TopDecl a) where
```
kosmikus, line 198: Let's give a name to this pair type. Something like `SectionedTopDecl`.

jl4-core/src/L4/Syntax.hs
```diff
@@ -12,13 +12,18 @@ import Data.Default
...
 
 data Name = MkName Anno RawName
   deriving stock (GHC.Generic, Eq, Show)
   deriving anyclass (SOP.Generic, ToExpr, NFData)
 
 data RawName =
     NormalName Text
+  | QualifiedName Text (NonEmpty Text)
+  -- ^ contains the actual name and a list of qualifiers, e.g.
+  -- foo.bar becomes @'QualifiedName' bar [foo]@
```
kosmikus, line 26: I see no good reason to flip the order around here, why not keep the qualifiers in front?

jl4-core/src/L4/TypeCheck/Types.hs
```diff
@@ -22,6 +23,7 @@ type Substitution = Map Int (Type' Resolved)
 data CheckEntity =
     KnownType Kind [Resolved] (TypeDecl Resolved)
   | KnownTerm (Type' Resolved) TermKind
+  | KnownSection (Section Resolved)
```
kosmikus, line 26: I am not sure I'd expect `Section Resolved` here.

I think I would expect something like `[CheckInfo]` here?

jl4-core/src/L4/TypeCheck/Types.hs
```diff
@@ -292,20 +294,86 @@ resolvedType n = do
         KnownType kind _resolved _tyDecl -> setAnnResolvedKindOfResolved kind n
         KnownTerm ty _term -> setAnnResolvedTypeOfResolved ty n
         KnownTypeVariable -> setAnnResolvedKindOfResolved 0 n
+        KnownSection _ -> n -- TODO: this is probably not what we want, maybe some internal error?
```
kosmikus, line 297: It looks strange to me that we couldn't be actually doing anything meaningful here?

jl4-core/src/L4/TypeCheck/Types.hs
```diff
@@ -292,20 +294,86 @@ resolvedType n = do
...
+      sectionInSection :: [[Section Resolved]] -> [Section Resolved]
+      -- NOTE: we have to treat this specially otherwise the recursive call will always
+      -- bottom out at an empty list of sections which makes the entire algorithm fail
+      sectionInSection [secs] = secs
+      sectionInSection (secs : secss) =
+        mapMaybe
+          (\sec@(MkSection _ mn _ _) -> do
+             u <- getUnique <$> mn
+             guard $ any (isTopLevelBindingInSection u) $ sectionInSection secss
+             pure sec
```
kosmikus, line 341: As I said above, it would feel more natural to me to look at this in terms of `[CheckInfo]` or similar rather than a `Section Resolved`.

Intuitively, we want the *type* of the section here, not the section itself.

kosmikus (commented):
Looks quite good in general. There are a few things I'd like to understand a bit better before this is merged, but it's plausible it can go in more or less as-is.

jl4-core/src/L4/Parser.hs
```diff
@@ -307,8 +309,31 @@ simpleName :: Parser (Epa Name)
...
+  let e = Epa
+        { original = map (either id fst) res
+        , trailingTokens = wsOrAnnotation.trailingTokens
+        , payload = QualifiedName n (q :| qs)
+        , hiddenClusters = wsOrAnnotation.hiddenClusters
+        }
+  pure $ MkName emptyAnno <$> e
+
+ where
+ tokOf p = token (\t -> (t,) <$> preview p (computedPayload t)) Set.empty
```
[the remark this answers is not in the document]
MangoIV: `tokOf` is just a helper function that does something similar that other parsers do. the just parse `preview ..` instead of the prism.

jl4-core/src/L4/Parser.hs
```diff
@@ -328,11 +353,7 @@ module' uri = do
   attachAnno $
     MkModule emptyAnno uri
       <$  annoLexeme_ spaceOrAnnotations
-      <*> annoHole
-          ((:)
-            <$> anonymousSection
-            <*> many section
-          )
+      <*> annoHole anonymousSection
```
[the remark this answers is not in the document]
MangoIV: I don't think so. Why would we?

jl4-core/src/L4/Print.hs
MangoIV: Makes sense, I'll do that

jl4-core/src/L4/Syntax.hs
MangoIV: I don't personally care, so I'll change that.

jl4-core/src/L4/TypeCheck/Types.hs
MangoIV: well a section doesn't have a type and this function is only useful for setting the type of that known thing

jl4-core/src/L4/TypeCheck/Types.hs
MangoIV: I don't have the same intuition. For me this is all about scoping, nothing much to do with the types, since sections cannot have types.

MangoIV: I agree that it's a bit weird since all of the constructs we had til now have had types and now it feels a bit intermingled but I think it makes sense for sections.

jl4-core/src/L4/Parser.hs
```diff
@@ -307,8 +309,31 @@ simpleName :: Parser (Epa Name)
...
+        , hiddenClusters = wsOrAnnotation.hiddenClusters
+        }
+  pure $ MkName emptyAnno <$> e
+
+ where
+ tokOf p = token (\t -> (t,) <$> preview p (computedPayload t)) Set.empty
+
 name :: Parser Name
-name = attachEpa (quotedName <|> simpleName) <?> "identifier"
+name = attachEpa (quotedName <|> try qualifiedName <|> simpleName) <?> "identifier"
```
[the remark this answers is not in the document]
MangoIV: I thought about this but since parser usually parse things and then some whitespace, there's not really a good way of reusing anything from there. The alternative is to just do it all in the lexer but then I felt a bit weird because we already lex everything that is contained in such a qualified name.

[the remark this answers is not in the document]
MangoIV: But maybe I'm misunderstanding your comment?

jl4-core/src/L4/Parser.hs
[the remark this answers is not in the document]
MangoIV: I don't really see why that would be problematic - the only difference is that I write `preview` once instead of multiple times. Maybe you have a more concrete suggestion though?

jl4-core/src/L4/Parser.hs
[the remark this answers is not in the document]
kosmikus: Because I think it might be a common use case to want to define some section and then have `#EVAL` directives that operate on the top-level. But I guess you can always start a *new* subsection to achieve more or less the same. And textual documents don't typically have a way to end sections either. So perhaps it's fine. Let's try for the time being.

jl4-core/src/L4/TypeCheck/Types.hs
kosmikus: Sections do have types!

jl4-core/src/L4/TypeCheck/Types.hs
kosmikus: Again, sections *have* types.

MangoIV: alright, I'll see if that works.

jl4-core/src/L4/Parser.hs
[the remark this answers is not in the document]
MangoIV: alright.

jl4-core/src/L4/TypeCheck/Types.hs
MangoIV: I actually need some information from the `Section` though, like its name etc. Idk how `CheckInfo` would improve the situation. I'm not seeing it (I also tried it out and I thought it would complicate things quite a lot and makes things harder to read?)

jl4-core/src/L4/TypeCheck/Types.hs
MangoIV: I have had a another look and I think this is for hovering type information. While I understand what you said wrt the types of sections, I don't think we want to show them in the user interface, right? So I guess we cannot do anything useful here, still, can we?

PR #354, opened 2025-04-15
provide prelude in data dir 

### LINKED ISSUE #345: how to import prelude?
Image

MangoIV:
- try to get prelude from data directory if it can't be found in the same directory and in the root directory
- move the libraries directory to jl4-core
- adjust test suite

resolved #345


jl4-lsp/jl4-lsp.cabal
MangoIV, line 1: revert this

PR #355, opened 2025-04-15
Allow forward references

fendor:


jl4-core/src/L4/TypeCheck/With.hs
```diff
@@ -17,6 +17,7 @@ import Data.Bifunctor
 data With e a =
     With e (With e a)
   | Plain a
+  deriving (Show, Generic)
```
fendor, line 20:
```suggestion
```

jl4-lsp/src/LSP/L4/Actions.hs
```diff
@@ -49,12 +49,9 @@ topDeclToCompletionItem name = \case
...
          _ -> CompletionItemKind_Constant
       }
-  KnownType kind _args tydec ->
+  KnownType kind _args _tydec ->
     Just (defaultTopDeclCompletionItem (typeFunction kind))
-      { CompletionItem._kind = Just $ case tydec of
-          RecordDecl {} -> CompletionItemKind_Struct
-          EnumDecl {} -> CompletionItemKind_Enum
-          SynonymDecl {} -> CompletionItemKind_Reference
+      { CompletionItem._kind = Just CompletionItemKind_Class
```
fendor, line 54: This was not really accurate in my opinion, this was just a type constructor of some kind, and it makes the impl more complicated, so I removed it :)

jl4-core/src/L4/TypeCheck.hs
```diff
@@ -187,18 +249,12 @@ runCheckUnique c e s =
   case runCheck c e s of
     [] -> error "internal error: expected unique result, got none"
     [(w, s')] -> (w, s')
-    _ -> error "internal error: expected unique result, got several"
+    _ -> error $ "internal error: expected unique result, got several"
```
fendor, line 252:
```suggestion
    _ -> error "internal error: expected unique result, got several"
```

MangoIV (approved):
I really don't like this but if you insist, we can do it. 

Not only does it change the behavior which makes it inconsistent with how this normally works, it also makes the environment blow up. 

It's not even significantly less code, especially if you think about having to mitigate the issue with the large environment.

PR #358, opened 2025-04-16
Viz: Refactor --  Inline FunDecl / Remove IRDecl, since we probably won't be Ladder-visualizing other sorts of Decls

ym-han:
This PR removes IRDecl, and thereby simplifies the code in the Ladder backend (as well as in the downstream consumers).

Viz still seems to show up.

image

image

### DIFF
--- a/jl4-lsp/src/LSP/L4/Viz/VizExpr.hs
+++ b/jl4-lsp/src/LSP/L4/Viz/VizExpr.hs
@@ -11,7 +11,7 @@ import GHC.Generics (Generic)
 import Optics
 
 newtype VisualizeDecisionLogicIRInfo = MkVisualizeDecisionLogicIRInfo
-  { program :: IRDecl -- TODO: change the fieldname once this becomes more stable
+  { program :: FunDecl -- TODO: change the fieldname once this becomes more stable
   }
   deriving newtype (Eq)
   deriving stock (Show, Generic)
@@ -23,8 +23,6 @@ data Name = MkName
   , label  :: Text     -- ^ Label to be displayed in the visualizer.
   }
   deriving (Show, Eq, Generic)
-
-type IRDecl = FunDecl
 
 -- TODO: Will worry about adding param type info later
 data FunDecl = MkFunDecl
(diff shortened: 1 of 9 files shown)


johsi-k:
No optical regressions found so far

jl4-lsp/src/LSP/L4/Viz/VizExpr.hs
```diff
@@ -74,21 +72,13 @@ instance HasCodec BoolValue where
...
-          <*> requiredField' "body" .= view _4
-
-      mkFunDecl (uid, name, params, body) = MkFunDecl uid name params body
+instance HasCodec FunDecl where
+  codec = object "FunDecl" $
+    typeField "FunDecl" MkFunDecl
+      <*> requiredField' "id"     .= view #id
+      <*> requiredField' "name"   .= view #fnName
+      <*> requiredField' "params" .= view #params
+      <*> requiredField' "body"   .= view #body
```
MangoIV, line 81: nice! I had already talked about that a few reviews ago, nice that you found a solution for that!

jl4-lsp/src/LSP/L4/Viz/VizExpr.hs
```diff
@@ -160,3 +150,13 @@ I haven't actually tried setting up such a automated test though.
 
 If that doesn't end up panning out, I will switch back to aeson.
 -}
+
+-------------------------------------------------------------
+-- Utils
+-------------------------------------------------------------
+{- | Adapted from
+https://github.com/NorfairKing/autodocodec/blob/e939442995debec6d0e014bfcc45449b3a2cb6e6/autodocodec-api-usage/src/Autodocodec/Usage.hs#L827
+-}
```
MangoIV, line 159: I don't think you need to add this link here, but it's fine if you do.

MangoIV: btw, idk if I had already asked this but have you checked the license of that code? norfairking sometimes has weird one off licenses that trap commercial users.

MangoIV (approved):
good changes. thank you.

jl4-lsp/src/LSP/L4/Viz/VizExpr.hs
ym-han: yeah was thinking of our convo too!

PR #356, opened 2025-04-16
Viz: Make it clearer in the types that the frontend evaluator only works with UBooleans

ym-han:
This adapts and generalizes a suggestion of @fendor 's.

(I've also continued doing a bit more on this front, e.g. removing the `Value` type from `eval/type.ts` and cleaning up related things a bit in a subsequent PR, on the `ym-han/issue-292` branch.)

PR #362, opened 2025-04-16
Viz: payload interfaces / Effect schemas for evaluating App expr(s) with args on backend

### LINKED ISSUE #292: Viz: Make proposal for start of simple protocol between visualizer and language server for evaluation and related things

ym-han:
Depends on #361  and #356 -- please review those first.

Closes #292 

### Summary

Key things:
* Add payload interfaces / Effect schemas for evaluating an App expr with args on backend, for both the request and response result
    * I'm thinking of using a custom method for the communication between the frontend and backend, but I haven't added that yet --- I want to ask fendor about it first.
    * This doesn't currently support *batched* requests, but it won't be hard to add that later. 
* Distinguish more explicitly between BoolValue and UBoolValue (and related things)
    * Evaluating an App on the backend uses BoolValues, as opposed to UBoolValues.

And some refactoring:
* Rename `message-types.ts` in the `viz-expr` package to the clearer `vscode-and-webview-messages.ts`
* Remove `Value` type from `eval/type.ts` (related to #356)

### Screenshots showing that this hasn't broken the ladder viz, at least not on my system:

image
image


jl4-lsp/src/LSP/L4/Viz/VizExpr.hs
```diff
@@ -19,7 +19,7 @@ newtype RenderAsLadderInfo = MkRenderAsLadderInfo
 type Unique = Int
 -- | TODO: Consider renaming this to something other than `Name`
 data Name = MkName
-  { unique :: Unique   -- ^ for checking whether, e.g., two BoolVar IRNodes actually refer to the same proposition.
+  { unique :: Unique   -- ^ for checking whether, e.g., two UBoolVar IRNodes actually refer to the same proposition.
```
ym-han, line 22: The changes in this file amount just to renaming BoolVar to UBoolVar etc to align with the more explicit distinction between BoolValues and UBoolValues on the frontend

johsi-k:
No optical regressions found so far

MangoIV (approved):

PR #368, opened 2025-04-21
Viz: Add App variant to IRExpr / VizExpr. For now, translate to IRExpr App on the Ladder backend only if dev mode flag is enabled and the App's function is a function from boolean(s) to a boolean

ym-han:
- [x] Add App schema / interface to viz-expr.ts
- [x] Refactor LSP / Ladder backend to pass info like the substitution from TC result and moduleUri along to Ladder backend
- [x] Add App to Ladder backend. For now, translate to IRExpr App only if the App's function is a function from boolean(s) to a boolean.
  - I use the Anno's TypeInfo for this. I'm not sure that's the best way to do it, but it also isn't important to get this right for now, since this is just a temporary restriction.
- [x] Hide the translate-to-App functionality in the Ladder backend under a dev mode flag for the time being (that is by default False). Will add AppLirNode and concrete UI in future PR(s).

If you enable `isDevMode` in Ladder.hs and reinstall the lsp, you'll see a placeholder 'not yet implemented' error for Apps of boolean functions (which demonstrates that the backend is able to pass the right kind of object to the frontend), e.g.

image

(Ignore the dots / hats in the buffer in the screenshot --- that's from another vscode extension that I personally use.)

### DIFF
--- a/jl4-lsp/src/LSP/L4/Viz/Ladder.hs
+++ b/jl4-lsp/src/LSP/L4/Viz/Ladder.hs
@@ -169,9 +169,9 @@ translateExpr False = go
           leafFromVizName (mkVizNameWith prettyLayout resolved)
 
         App appAnno fnResolved args -> do
-          fnOfAppisFnFromBooleanToBooleans <- and <$> traverse hasBooleanType (appAnno : map getAnno args)
+          fnOfAppIsFnFromBooleansToBoolean <- and <$> traverse hasBooleanType (appAnno : map getAnno args)
           -- for now, only translating App of boolean functions to V.App
-          if isDevMode && fnOfAppisFnFromBooleanToBooleans
+          if isDevMode && fnOfAppIsFnFromBooleansToBoolean
             then
               V.App
                 <$> getFresh
--- a/jl4-lsp/src/LSP/L4/Viz/Ladder.hs
+++ b/jl4-lsp/src/LSP/L4/Viz/Ladder.hs
@@ -168,8 +168,8 @@ translateExpr False = go
         App _ resolved [] ->
           leafFromVizName (mkVizNameWith prettyLayout resolved)
 
-        fnApp@(App _ fnResolved@(getOriginal -> fnName) args) -> do
-          fnOfAppisFnFromBooleanToBooleans <- and <$> traverse exprHasBooleanType (fnApp : args)
+        App appAnno fnResolved args -> do
+          fnOfAppisFnFromBooleanToBooleans <- and <$> traverse hasBooleanType (appAnno : map getAnno args)
           -- for now, only translating App of boolean functions to V.App
           if isDevMode && fnOfAppisFnFromBooleanToBooleans
             then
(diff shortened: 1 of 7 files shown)


jl4-lsp/src/LSP/L4/Viz/Ladder.hs
```diff
@@ -185,3 +225,21 @@ mkVizNameWith :: (Name -> Text) -> Resolved -> V.Name
...
+exprHasBooleanType _ = pure False
+
+-- | Returns True iff the Type Resolved is that of a L4 BOOLEAN
+isBooleanType :: Type' Resolved -> Viz Bool
+isBooleanType ty = do
+  type' <- getExpandedType ty
+  pure $ case type' of
+    TyApp _ (Ref _ uniq _) [] ->
+      uniq == TC.booleanUnique
+    _ -> False
```
MangoIV, line 245: all of these are "pure" functions, they don't change the state. Normally you'd not put the environment in the state to avoid things like this. I think it's fine here though

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
```diff
@@ -133,12 +163,22 @@ translateExpr False = go
...
+        -- TODO: Should handle BoolLits differently too
         -- 'var'
         App _ resolved [] ->
           leafFromVizName (mkVizNameWith prettyLayout resolved)
-        -- TODO: Will be replacing this temporary version with a variant for App on the frontend
-        App _ (getOriginal -> MkName _ fnName) args ->
-          leaf "" $ Text.unwords (rawNameToText fnName : (prettyLayout <$> args))
+
+        fnApp@(App _ fnResolved@(getOriginal -> fnName) args) -> do
+          fnOfAppisFnFromBooleanToBooleans <- and <$> traverse exprHasBooleanType (fnApp : args)
```
MangoIV, line 172: I think it would've been better to pass in the annos directly here.

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
```diff
@@ -133,12 +163,22 @@ translateExpr False = go
...
 
+        -- TODO: Should handle BoolLits differently too
         -- 'var'
         App _ resolved [] ->
           leafFromVizName (mkVizNameWith prettyLayout resolved)
-        -- TODO: Will be replacing this temporary version with a variant for App on the frontend
-        App _ (getOriginal -> MkName _ fnName) args ->
-          leaf "" $ Text.unwords (rawNameToText fnName : (prettyLayout <$> args))
+
+        fnApp@(App _ fnResolved@(getOriginal -> fnName) args) -> do
```
MangoIV, line 171: I think this is goo complicated. there's not much point in using a view pattern that doesn't do any matching. Also if you see the comment above, you should probably simplify the at pattern.

MangoIV (approved):
I think semantically this is fine.

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
ym-han: Yeah, I thought about using a Reader, but it also felt like it might be overkill.

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
ym-han: Good point, thanks!

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
ym-han: This does help make it more readable, thank you!

PR #378, opened 2025-04-27
VSCode ext: Refactor -- Improve some out-of-date names

ym-han:
Depends on #377 -- please review that first

- VSCode: Rename the project from `jl4-lsp-client` to the more precise `jl4-vscode`. 
  - This will allow us to use the more general name for a package that will be shared among all lsp clients. 
- Rename `vscode/src/viz.ts` to the clearer `vscode/src/webview-panel.ts`


MangoIV (approved):
lgtm!

PR #379, opened 2025-04-28
Init `ts-shared/jl4-lsp-client` TS project with `custom-protocol.ts` for our custom LSP extensions (tho currently only contains the `EvalAppRequest` type)

ym-han:
Closes #380 

Refactoring:
* Misc refactor: Improve viz-expr project's eslint config (don't need the additional svelte eslint stuff)
* Refactor: Rename `EvalAppInfo` to the more LSP-y `EvalAppRequestParams`

Init a shared LSP client package:
* Init the `ts-shared/jl4-lsp-client` TS project / package, to be used by the various LSP clients / IDE extensions
* Add a `custom-protocol.ts` that's meant to collect all the custom LSP extensions, though for now there's just the `EvalAppRequest` type

PR #383, opened 2025-04-29
Client RPC: Add a `L4LanguageClient` interface and wrapper object implementations that wrap over the language client of the respective IDEs. Make a `L4RpcRequestType` for all client-side RPC and export a ctor for that. Rename `jl4-lsp-client` to the more general `jl4-client-rpc`

ym-han:
Depends on #379 -- please review that first.

If you see a better way to structure the projects or to do the interfaces, please do point that out.

This PR is part of an effort to come up with a way for the frontend to (indirectly) send requests to the LSP server, and to better organize the client-side RPC.

The key things this PR does include:
- LSP Protocol (and client-side RPC in general): Make a `L4RpcRequestType` and export a ctor for that, instead of exposing implementational details re what exact request type we are using
   - Add conversion helpers for converting between L4RpcRequestType and the webview messenger request type
- Add `L4LanguageClient` interface to the `jl4-lsp-client` package.
   - This uses a lower-level `sendRequest` method (we'll in effect wrap over it with higher-level methods in, e.g., the interface that the frontend will use).
- Add simple classes implementing that for both VSCode and Monaco; update the extensions accordingly to use the wrapper object.
- Rename `jl4-lsp-client` to the more general `jl4-client-rpc` (since we've moved things like messages between vscode ext and vscode webview there as well); update the package description

And misc refactoring / infra:
- rename LspResult to LspResponse
- Add misc docstrings
- Re-export EvalAppRequestParams, EvalAppResult from jl4-client-rpc
- Add await for some async functions in jl4-web's +page.svelte
- Infra: Improve `files` entry of viz-expr's package.json
- Infra: Improve jl4-client-rpc and viz-expr's tsconfig.json
- Infra: Add vscode-jsonrpc dep; add clean script for `jl4-lsp-client` and `viz-expr`. **Clean before building.**


MangoIV (approved):
lgtm

PR #388, opened 2025-04-29
Add semantic tokens based on the typechecked AST

### LINKED ISSUE #170: Improve syntax highlighting and add syntax highlighting based on typecheck result
Syntax highlighting is currently only based on the lexer and parse results.
Even then, we are not highlighting to the best of our knowledge, take:

Image

In this context, we *know* that `Tree a` must both be types, but we are highlighting them as values.

Further, we should provide syntax highlighting based on the typecheck result of the AST which could allow us to provide more accurate highlighting in at least the following cases:

* selectors
* functions vs data constructors
* parameters
* assumes

fendor:
image
We add a new 'Rule' `TypeCheckedSemanticTokens` which generates
'SemanticTokens' based on 'Module Resolved'.
These tokens are merged with the result of `ParserSemanticTokens` and
`LexerSemanticTokens`, to provide the most accurate and helpful syntax
highlighting.

We have to very careful when modifying or generating AST nodes during the typecheck
phase to not accidentally reuse `Anno`s of `Name`s or similar.

For example, given this L4 code:

```
GIVEN a
foo MEANS a
```

The parsed AppForm is 'MkAppForm ... "foo" []'. But during typechecking,
we normalise this, and add the "a" from the "TypeSig" to the "AppForm",
resulting in an AST node like: 'MkAppForm ... "foo" ["a"]'.
In this situation, we need to make sure we remove the 'Anno' associated
with the "a", otherwise, during syntax highlighting, we will write out
the "a" after "foo" in the semantic tokens, leading to incorrect syntax
highlighting.

Closes #170


jl4-lsp/src/LSP/Core/Types/Options.hs
```diff
@@ -117,7 +117,14 @@ defaultIdeOptions = IdeOptions
...
     ,optRunSubset = True
-    ,optPreservedKeys = \_ -> PreservedKeys HashSet.empty
+    ,optPreservedKeys = \_ -> PreservedKeys $ HashSet.fromList
+        -- always preserved
+        [ typeOf GetFileExists
+        , typeOf GetModificationTime
+        , typeOf IsFileOfInterest
+        , typeOf GetClientSettings
+        , typeOf AddWatchedFile
+        ]
```
MangoIV, line 127: what does this do?

jl4-lsp/src/LSP/L4/Rules.hs
```diff
@@ -741,3 +723,56 @@ rangeOfResolveWarning = \case
...
+--
+-- TODO: this function is inefficient, as all phases are always executed.
+-- It would be better if we could tell whether the result is actually stale and then
+-- we won't run later phases.
+-- E.g. @NormalizedUri -> Action (Maybe ([SemanticToken], Maybe PositionMapping))@
+semanticTokensUsing ::
+  [NormalizedUri -> Action (Maybe ([SemanticToken], PositionMapping))] ->
+  (NormalizedUri -> Action [SemanticToken])
+semanticTokensUsing phases uri = do
+  semTokens <- traverse ($ uri) phases
```
MangoIV, line 756: why not apply the function to the uri before you hand it to this function? i.e. 
`[Action (Maybe (..)] -> Action [SemanticToken]`?

jl4-lsp/src/LSP/L4/Rules.hs
```diff
@@ -741,3 +723,56 @@ rangeOfResolveWarning = \case
...
+  mergeSameLengthTokens [] bs = bs
+  mergeSameLengthTokens as [] = as
+  mergeSameLengthTokens (a : as) (b : bs) = case compare a.start b.start of
+    -- a.start == b.start
+    -- Same token, only print one
+    EQ -> a : mergeSameLengthTokens as bs
+    -- a.start < b.start
+    LT -> a : mergeSameLengthTokens as (b : bs)
+    -- a.start > b.start
+    GT -> b : mergeSameLengthTokens (a : as) bs
```
MangoIV, line 778: the part that contains the actual logic is  a pure function. I think that fact should be reflected in the code. 

Tbh I also don't understand why this is correct. Doesn't merging in that way require that the tokens are ordered by length? Is that the case?

jl4-lsp/src/LSP/L4/SemanticTokens.hs
```diff
@@ -214,3 +236,98 @@ instance ToSemTokens () PosToken PosToken where
...
+      Def _ n -> modifier entityInfo $ withModifier defVar $ toSemTokens n
+      Ref orig _ _ -> modifier entityInfo $ toSemTokens orig
+      OutOfScope _ n -> modifier entityInfo $ toSemTokens n
+    where
+      uniq = getUnique r
+      modifier entityInfo =
+        case Map.lookup uniq entityInfo of
+          -- Strictly speaking, this is an internal invariant violation.
+          -- However, we likely shouldn't care during syntax highlighting.
+          Nothing -> id
```
MangoIV, line 257: i think we should put an `assert` on those cases.

jl4-lsp/src/LSP/L4/SemanticTokens.hs
```diff
@@ -43,13 +45,13 @@ standardTokenType = \case
 simpleTokenType :: PosToken -> Maybe SemanticTokenTypes
 simpleTokenType t = standardTokenType (posTokenCategory t.payload)
 
-parameterType :: PosToken -> Maybe SemanticTokenTypes
-parameterType t = case posTokenCategory t.payload of
+identIsParameter :: PosToken -> Maybe SemanticTokenTypes
+identIsParameter t = case posTokenCategory t.payload of
```
MangoIV, line 49: this has been renamed, but it appears to be unused?

jl4-lsp/src/LSP/L4/SemanticTokens.hs
```diff
@@ -128,13 +145,20 @@ instance ToSemToken PosToken where
...
+  toSemTokens (MkModule ann _uri sects) =
+    traverseCsnWithHoles ann [toSemTokens sects]
+deriving anyclass instance ToSemTokens Context PosToken (TopDecl Name)
+
 instance ToSemTokens Context PosToken (Section Name) where
   toSemTokens (MkSection ann name maka decls) =
     traverseCsnWithHoles ann
-      [ withTokenType nameIsDirective $ toSemTokens name
-      , withTokenType nameIsDirective $ toSemTokens maka
+      [ withTokenType identIsDirective $ toSemTokens name
```
MangoIV, line 160: I don't understand why it is called that. Afaiu, this doesn't check if the given identifier is a directive, it just checks whether it is an identifier.

jl4-lsp/src/LSP/L4/SemanticTokens.hs
```diff
@@ -183,8 +203,10 @@ deriving anyclass instance ToSemTokens Context PosToken (Decide Name)
 deriving anyclass instance ToSemTokens Context PosToken (Aka Name)
 deriving anyclass instance ToSemTokens Context PosToken (TypeDecl Name)
 deriving anyclass instance ToSemTokens Context PosToken (NamedExpr Name)
-deriving anyclass instance ToSemTokens Context PosToken (Branch Name)
-deriving anyclass instance ToSemTokens Context PosToken (Pattern Name)
+instance ToSemTokens Context PosToken (Branch Name) where
+  toSemTokens = genericToSemTokens
+instance ToSemTokens Context PosToken (Pattern Name) where
+  toSemTokens = genericToSemTokens
```
MangoIV, line 209: this change seems unnecessary.

jl4-lsp/src/LSP/SemanticTokens.hs
```diff
@@ -45,7 +45,7 @@ class ToSemTokens c t a | a t -> c where
   toSemTokens = genericToSemTokens
 
 genericToSemTokens :: forall a t c .
-    (SOP.Generic a, All (AnnoFirst a (ToSemTokens c t)) (Code a), HasAnno a, ToSemToken t, AnnoToken a ~ t) =>
+    (HasCallStack, SOP.Generic a, All (AnnoFirst a (ToSemTokens c t)) (Code a), HasAnno a, ToSemToken t, AnnoToken a ~ t) =>
```
MangoIV, line 48: this change seems unnecessary

jl4/tests/SemanticTokens.hs
```diff
@@ -0,0 +1,71 @@
...
+    lexerTokens <- Shake.use Rules.LexerSemanticTokens uri
+    parserTokens <- Shake.use Rules.ParserSemanticTokens uri
+    checkedTokens <- Shake.use Rules.TypeCheckedSemanticTokens uri
+    pure (lexerTokens, parserTokens, checkedTokens)
+  let
+    output_ :: Text -> Maybe [SemanticToken] -> Text
+    output_ phase mToks =
+      "Phase: " <> phase <> "\n" <>
+      case mToks of
+        Nothing -> "No Tokens for this phase"
```
MangoIV, line 42: that's not what this means, it means that the entire thing failed. I think we should rather assert that this is `Just` up top and then just not have this case here.

jl4/examples/lsp/semantic-tokens/annotations.l4
MangoIV, line 1: can we do without duplicating all these files? e.g. with symlinks?

MangoIV (approved):

jl4-lsp/src/LSP/L4/Rules.hs
fendor: I just thought it looks nicer when calling :shrug:

jl4-lsp/src/LSP/Core/Types/Options.hs
fendor: Avoid garbage collection of important keys

This avoids overly-eager garbage collection of results for certain keys.
All the keys we preserve are taken from HLS.

This is mostly me sneakily adding orthogonal commits to PRs out of laziness.

jl4-lsp/src/LSP/L4/SemanticTokens.hs
```diff
@@ -214,3 +236,98 @@ instance ToSemTokens () PosToken PosToken where
...
+deriving anyclass instance ToSemTokens EntityInfo PosToken (Aka Resolved)
+deriving anyclass instance ToSemTokens EntityInfo PosToken (TypeDecl Resolved)
+deriving anyclass instance ToSemTokens EntityInfo PosToken (NamedExpr Resolved)
+deriving anyclass instance ToSemTokens EntityInfo PosToken (Branch Resolved)
+deriving anyclass instance ToSemTokens EntityInfo PosToken (Pattern Resolved)
+deriving anyclass instance ToSemTokens EntityInfo PosToken (TypeSig Resolved)
+deriving anyclass instance ToSemTokens EntityInfo PosToken (GivethSig Resolved)
+deriving anyclass instance ToSemTokens EntityInfo PosToken (GivenSig Resolved)
+deriving anyclass instance ToSemTokens EntityInfo PosToken (Directive Resolved)
+deriving anyclass instance ToSemTokens EntityInfo PosToken (Import Resolved)
```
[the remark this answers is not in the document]
MangoIV: because deriving anyclass instance Foo is equivalent to instance Foo, so unnecessarily verbose and if we want to give an explicit impl, the diff is noisier, too. 

So if we’re having this diff once for moving anyway, I think we should instead use empty instance decls

jl4-lsp/src/LSP/L4/SemanticTokens.hs
fendor: This comment is out of date, many `Resolved` are not part of the `entityInfo`, for example `type variables`.

jl4-lsp/src/LSP/Core/Types/Options.hs
MangoIV: No this is fine, I just wanted to know what it does. Thanks!

jl4-lsp/src/LSP/L4/Rules.hs
MangoIV: Oh yeah start is what I meant! Yeah then this makes sense! Maybe you should add a comment „tokens are ordered by the length of the starting symbol“

jl4-lsp/src/LSP/L4/SemanticTokens.hs
fendor: The `` § `var` `` the `var` is backed by a `Name` which is backed by a `TIdentifier`. To highlight this name as an `Directive`, we need to say each `TIdentifier` is to be highlighted as an `SemanticToken_Directive`.

jl4-lsp/src/LSP/L4/SemanticTokens.hs
fendor: Thanks, that was only for debugging!

jl4-lsp/src/LSP/SemanticTokens.hs
fendor: Also, debugging remnant, thanks!

jl4/tests/SemanticTokens.hs
fendor: No, for example when a file doesn't typecheck, then this phase may reasonably not produce any tokens.
None of the tests currently test the case of partially correct files, but I would like to add some.

MangoIV: But doesn’t this already return Nothing when any of the phases fail?

jl4/examples/lsp/semantic-tokens/annotations.l4
fendor: They are not duplicated right now, most of these files have been modified intentionally to cover as many edge cases for syntax highlighting as possible. We could try to deduplicate them, but deduplication in test data doesn't feel necessary to me.

MangoIV (approved):
consider addressing my nitpicks, but if not, I'm good with this as is.

fendor:
I still have to fix the regression, so this is not ready to merge, yet :(

MangoIV:
Oh yeah sure

jl4-core/src/L4/Utils/IntervalMap.hs
```diff
@@ -0,0 +1,53 @@
...
+  let
+    r = IV.search sp info
+  in
+    case r of
+      [] -> Nothing
+      _ -> Just (first (intervalToRange0 nuri) $ maximumBy cmp r)
+ where
+  cmp (a, _) (b, _) =
+    (b.high `compare` a.high)
+      <> (a.low `compare` b.low)
```
MangoIV, line 41: shouldn't this be just `listToMaybe $ sortOn (\i -> i.hi - i.lo) $ search sp info`? What additionaly thing does this do

jl4-lsp/src/LSP/L4/SemanticTokens.hs
```diff
@@ -216,3 +237,93 @@ instance ToSemTokens () PosToken PosToken where
...
+--- ---------------------------------------------------------------------------
+-- Semantic Tokens for typechecked AST
+--- ---------------------------------------------------------------------------
+
+instance ToSemTokens () PosToken Resolved where
+  toSemTokens r = do
+    case r of
+      Def _ n -> modifier (infoOf n) $ toSemTokens n
+      Ref n _ _ -> modifier (infoOf n) $ toSemTokens n
+      OutOfScope _ n -> modifier (infoOf n) $ toSemTokens n
```
MangoIV, line 250:
```suggestion
    let n = getActual r in modifier (infoOf n) $ toSemTokens n
```

jl4-lsp/src/LSP/L4/SemanticTokens.hs
```diff
@@ -216,3 +237,93 @@ instance ToSemTokens () PosToken PosToken where
...
+                Nothing -> id
+          MyKnownTypeVariable ->
+            withTokenType identIsTypeVar
+
+      basedOnType = \case
+        Type{} -> id
+        Fun{} -> withTokenType identIsFunction
+        TyApp{} -> id
+        Forall{} -> id
+        InfVar{} -> id
```
MangoIV, line 273:
```suggestion
      highlightFunction = \ case
        Fun{} -> withTokenType identIsFunction
        _ -> id
```

jl4-lsp/src/LSP/SemanticTokens.hs
```diff
@@ -34,7 +34,7 @@ runSemanticTokensM semTokenCtx a = Except.runExcept $ ReaderT.runReaderT (toSemT
 --
 -- We might want to override some functionality here. We should perhaps
 -- try to find another way to do this.
-class ToSemTokens c t a | a t -> c where
+class ToSemTokens c t a | a c -> t where
```
MangoIV, line 37: is this a bug fix? Or how is this change in semantics justified? Does this actually encode a rule for the type class or is this just for slightly improving type inference when inferring the instances?

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
```diff
@@ -324,8 +324,8 @@ mkVizNameWith printer (getUniqueName -> (uniq, name)) =
 ------------------------------------------------------
 
 hasBooleanType :: Anno_ t Extension -> Viz Bool
-hasBooleanType (Anno {extra = Extension {resolvedInfo = Just (TypeInfo ty _)}}) =
-  isBooleanType ty
+-- hasBooleanType (Anno {extra = Extension {resolvedInfo = Just (TypeInfo ty _)}}) =
+--   isBooleanType ty
```
MangoIV, line 328: this looks wrong

jl4-lsp/src/LSP/L4/Rules.hs
```diff
@@ -739,3 +727,56 @@ rangeOfResolveWarning = \ case
...
+          Just newPos -> Just (t & #start .~ newPos)
+    )
+    semTokens
+
+-- | @'semanticTokensUsing' phases@
+--
+-- Helper function for defining multi-phase semantic syntax highlighting.
+--
+-- Each phase can produce '[SemanticToken]'s and 'PositionMapping' if the result is outdated.
+-- Tokens obtained from earlier phases take precedence over tokens from later phases.
```
MangoIV, line 750: shouldn't it be the other way around?

jl4-lsp/src/LSP/L4/Rules.hs
```diff
@@ -739,3 +727,56 @@ rangeOfResolveWarning = \ case
...
+--
+-- TODO: this function is inefficient, as all phases are always executed.
+-- It would be better if we could tell whether the result is actually stale and then
```
MangoIV, line 753: is there a reason why we cannot? can't you check whether the rule currently succeeds and then use the stale result if it doesn't?

MangoIV (commented):
There are still some unaddressed things from my last review, I think.

Looks largely good to me. I don't think it's very maintainable to have the typechecker explicitly change the state to add information to the maps - that's something we should strive to avoid, if at all possible.

jl4-lsp/src/LSP/L4/Rules.hs
fendor: I came up with an artificial reason! It is easier to misuse an API (`[Action (Maybe (..)] -> Action [SemanticToken]`) as there is no guarantee that the `Action` has been called with the same `NormalizedFilePath` :thinking: 

Ok, joking aside, I don't feel too strongly about it, but I do prefer this a little bit.

jl4/tests/SemanticTokens.hs
fendor: No, the output is `moutput :: (Maybe [SemanticToken], Maybe [SemanticToken],
 Maybe [SemanticToken])`, I want to explicitly know when a specific phase fails and record it in the test file.

jl4-lsp/src/LSP/L4/Rules.hs
MangoIV: Yeah I'm sure it's fine :)

jl4-lsp/src/LSP/SemanticTokens.hs
fendor: This is only for type inference, looks like we don't need it at all.

jl4-lsp/src/LSP/L4/Rules.hs
fendor: I think it is possible, but requires a new primitive. A combination of `use` and `useWithStale` also works, but is likely a bit more expensive.

fendor: I implemented it in terms of `use` and `useWithStale`.

jl4-core/src/L4/Utils/IntervalMap.hs
fendor: `SrcPos` does not implement `Num`, otherwise it is the same

fendor: Reimplemented just like you proposed.

PR #389, opened 2025-04-30
TS infra: Add `npm run clean` script to TS projects that produce build artifacts, and as a turbo task. Modify build scripts to clean before building.

ym-han:


MangoIV (approved):

MangoIV:
that's nice. I had missed this when working on the ts.

ym-han:
> that's nice. I had missed this when working on the ts.

Yeah I should have done this earlier. Better late than never!

ym-han:
Oh but just to clarify, this doesn't remove installation-related stuff like node_modules. It only removes things that are created by npm run build.

Issue #396, opened 2025-05-02
Viz: Have client-server communication re ladder viz use LSP.VersionedTextDocumentIdentifier instead of just the uri


ym-han:
Copy of the stuff in the PR text:

 
- VersionId refactoring: Rename VersionId to the clearer VersionedDocId.
   Simplify: Just use number for the version, instead of a pseudo 'integer'
   Simplify: Remove the $type: VersionedDocId tag from VersionedDocId
- Backend: Send the VersionedTextDocId (or something roughly iso to that type) along to the extensions / lang clients when making the viz expr. Augment RenderAsLadderInfo with VersionedDocId.
- Lang clients: Update VSCode extension's and Monaco's didChange handler to use the versioned text doc id instead of just the uri

PR #397, opened 2025-05-02
Viz: Have client-server communication  re ladder viz use LSP's VersionedTextDocId instead of just the uri (issue #396)

ym-han:
Closes issue #396 

**Depends on PR #395  --- please review that first**

The main changes:

- VersionId refactoring: Rename `VersionId` to the clearer `VersionedDocId`. 
Simplify: Just use `number` for the version, instead of a pseudo 'integer'
Simplify: Remove the `$type: VersionedDocId` tag from `VersionedDocId`
- Backend: Send the VersionedTextDocId (or something roughly iso to that type) along to the extensions / lang clients when making the viz expr. Augment `RenderAsLadderInfo` with `VersionedDocId`.
- Lang clients: Update VSCode extension's and Monaco's `didChange` handler to use the versioned text doc id instead of just the uri

Unrelated refactoring:
* **jl4-web**: Wrap the persistSession fn call in a try/catch so the rest of the didChange callback will run even if the http calls in persistSession don't succeed. (This makes it easier to test locally.)
* Add some docstrings / comments here and there

Note on future plans: I'll factor out the Viz monad in the Ladder backend to a separate file at some point (and when I do that, I will only export the 'smart' ctor for VizEnv)

Screenshot of how the versioned text doc id seems to be being passed along (and how it seems to be updated in the way we'd like on changes to the doc):
image


jl4-lsp/src/LSP/L4/Viz/VizExpr.hs
```diff
@@ -133,7 +141,22 @@ instance HasCodec RenderAsLadderInfo where
...
+          <$> requiredField' "verTextDocId" .= view #verTextDocId
+          <*> requiredField' "funDecl"      .= view #funDecl
+
+instance HasCodec LSP.VersionedTextDocumentIdentifier where
+  codec =
+    named "VersionedDocId" $
+      object "VersionedDocId" $
+        LSP.VersionedTextDocumentIdentifier
+          <$> requiredField' "uri"     .= view #_uri
+          <*> requiredField' "version" .= view #_version
```
MangoIV, line 153: I think it would be better to implement this instance directly for VersionedDocId since 
1. its FromJSON/ToJSON instances doesn't cohere with this instance
2. it's an orphan that we could skip as we have a newtype wrapper anyway 

In general I think it would be less maintenance overhead if we just used *its* ToJSON/FromJSON instance - wouldn't that also fit better with the `LSP`? (because it would use the encoding that the LSP expects anyway, I guess?)

jl4-lsp/src/LSP/L4/Actions.hs
```diff
@@ -139,7 +140,7 @@ visualise mtcRes (getRecVis, setRecVis) uri msrcPos = do
   case mdecide of
     Nothing -> pure (InR Null)
     Just (decide, simp, substitution) ->
-      case Ladder.doVisualize decide (Ladder.MkVizEnv (toNormalizedUri uri) substitution simp) of
+      case Ladder.doVisualize decide (Ladder.mkVizEnv verTextDocId substitution simp) of
```
MangoIV, line 143: I think I would have personally (so no need to change it) not have introduced this smart constructor, because it doesn't "do" anything, it would've been easier to read to just have `Ladder.MkVizEnv verTextDocId._uri ...`

jl4-lsp/app/LSP/L4/Handlers.hs
```diff
@@ -276,17 +276,18 @@ handlers recorder =
...
-          use_ TypeCheck (toNormalizedUri uri)
+          use_ TypeCheck (toNormalizedUri verTextDocId._uri)
 
         let
           mkCodeLens srcPos simplify = CodeLens
             { _command = Just Command
               { _title = t simplify
               , _command = "l4.visualize"
-              , _arguments = Just [Aeson.toJSON uri, Aeson.toJSON (Generically srcPos), Aeson.toJSON simplify]
+              , _arguments = Just [Aeson.toJSON verTextDocId, Aeson.toJSON (Generically srcPos), Aeson.toJSON simplify]
```
MangoIV, line 290: do I understand correctly that this doesn't even use the custom versioned text doc ToJSON/FromJSON instances?

MangoIV (changes requested):

jl4-lsp/app/LSP/L4/Handlers.hs
ym-han: This uses the VersionedDocId instances (more precisely: the To/FromJSON instances produced by Autodocodec from the Codec instances in VizExpr).

MangoIV: Are you sure? At least the parsing doesn’t.

jl4-lsp/src/LSP/L4/Viz/VizExpr.hs
ym-han: Yeah it'd definitely be better to implement it directly for VersionedDocId (assuming we want to use VersionedDocId / some kind of custom type that's iso to the LSP.VersionedTextDocumentIdentifier). This just happened to be more convenient, but I probably should just implement it directly.

As for VersionedDocId vs LSP.VersionedTextDocumentIdentifier: I'm leaning towards VersionedDocId, but I don't feel strongly about this. I'll probably just go with what would require the least effort to change, unless you feel strongly in favor of using LSP.VersionedTextDocumentIdentifier.

MangoIV: I think we misunderstand each other - VersionedTextDocumentIdentifier already has an encoding specified in the protocol, right? It seems redundant to invent a custom interface on top of this.

jl4-lsp/app/LSP/L4/Handlers.hs
ym-han: I'm not sure what behavior you are seeing, but could it be because the current VersionedDocId instance basically piggybacks (via `dimapCodec`) on the orphan instance for LSP.VersionedTextDocumentIdentifier?

MangoIV: No. This is direct parseJSON @VersionedTextDocumentIdentifier which is an instance that is already part of the library (even without the Codec instance)

ym-han: You are right, good catch. I had forgotten about

```
verTextDocId <- liftIO $ runAction "codeLens.getVersionedTextDoc" ide $
          FileStore.getVersionedTextDoc $ params ^. J.textDocument
```

So, yeah, I should either use LSP. VersionedTextDocumentIdentifier or convert explicitly to the custom VersionedDocId

MangoIV: At least iiuc

jl4-lsp/src/LSP/L4/Viz/VizExpr.hs
ym-han: > I think we misunderstand each other - VersionedTextDocumentIdentifier already has an encoding specified in the protocol, right? It seems redundant to invent a custom interface on top of this.

So, if I'm understanding correctly, there are two different questions here:

1. Whether to use VersionedTextDocumentIdentifier directly or a custom type that's isomorphic to it like VersionedDocId
2. Instances: if we do use LSP.VersionedTextDocumentIdentifier, whether to just use the instance that comes with it. By using the instance, I mean: using `codecViaAeson` from Autodocodec to use the LSP.VersionedTextDocumentIdentifier's From/ToJSON instance to produce an Autodocodec codec.

I don't have a very strong opinion re 1: I'm happy to just refactor it to use LSP.VersionedTextDocumentIdentifier.

Re 2: the autodocodec docs suggest that using `codecViaAeson` is not a good idea, 

> You will only want to use this if you cannot figure out how to produce a [JSONCodec](https://hackage.haskell.org/package/autodocodec-0.4.0.0/docs/Autodocodec.html#t:JSONCodec) for your type.

> Note that this will not have good documentation because, at a codec level, it's just parsing and rendering a [Value](https://hackage.haskell.org/package/aeson-2.2.3.0/docs/Data-Aeson-Types.html#t:Value).

but it probably also doesn't matter for a type as simple as this.

jl4-lsp/app/LSP/L4/Handlers.hs
ym-han: Yes, I hadn't been 100% sure that the design intent of our VersionedDocId would be exactly the same as that of VersionedTextDocumentIdentifier, but yeah, now that I think more about it, let's just reuse it directly!

It's not going to make much of a difference on the client side, but it would indeed be cleaner on the server side

jl4-lsp/src/LSP/L4/Viz/VizExpr.hs
MangoIV: > whether to just use the instance that comes with it.

Mind that this only matters if you want to wrap it somewhere. For encoding/decoding To/From JSON are obviously already enough. So if we’re just skipping the codec instances and the wrapper, we should be golden. 

Plus, the usage here would already be correct, as from its perspective, the codec instances are dead code anyway, atm.

jl4-lsp/app/LSP/L4/Handlers.hs
ym-han: OK I've pushed the following changes:

* Refactor Ladder backend to use LSP.VersionedTextDocumentIdentifier directly instead of a wrapper over it
* Refactor monaco and vscode extensions to use the code2Protocol converter from `vscode-languageclient/lib/common/codeConverter.js` to get a VersionedTextDocIdentifier from the Code/Monaco.TextDocument. I.e., the conversion code has been refactored to just re-use their glue code.

The one other thing that one might want to change in this vein is the EvalAppRequestParams on the frontend: it's still using the VersionedDocId Effect schema instead of the VSCode/LSP VersionedTextDocIdentifier. I'm just holding off changing that for now, because that would probably also require not using Effect schemas at all for EvalAppRequestParams, and I would prefer not to have to think through / make that decision right now.

Thanks for pointing this out -- I definitely tend to reach too often for wrapper types

ym-han: I've thought a bit more about it. I would prefer to minimize further changes in the short term, but I have a concrete argument now for why we shouldn't just directly reuse LSP's VersionedTextDocumentIdentifier (and instead have something isomorphic / structurally the same as that type).

The argument is this. There's a subtle difference in the design intent of the Versioned*Text*DocumentIdentifier and our VersionedDocId. As the name of the former suggests, this is a type that seems specific to *text* documents, at least in design intent (even if this doesn't make a concrete difference in the fields etc);
 this impression is further corroborated by how the LSP spec also has, e.g., a distinct [VersionedNotebookDocumentIdentifier](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#versionedNotebookDocumentIdentifier) type that is structurally the same as the VersionedTextDocumentIdentifier.

But whether the source program was a text document or notebook or something else altogether, is *not* something that the *Ladder visualizer* cares about. (Or at least, that was a deliberate design decision I had been trying out --- so, e.g, the Ladder visualizer does not know about things like source positions or ranges.) That is, the `VersionedDocId` type is meant to be agnostic about the format of the original source document; and for that reason, is not something that should evolve in lockstep with changes to VersionedTextDocumentIdentifier (e.g. if they decide to one day add very text-doc-specific fields to LSP's VersionedTextDocumentIdentifier that don't matter to the visualization, we would not want to also forward those fields to the visualizer).

I'm not 100% sure about this; just wanted to try to make the inchoate worry I had more concrete.

And more importantly, it probably doesn't make sense to make a lot more changes to this in the short term

MangoIV: that seems reasonable but maybe we also want to think about premature complication ~ just b/c something might come up in the future, we shouldn't complicate it now, if we are sure we don't block any roads prematurely, but *not* adding something, i.e. just choosing a subset of what would be implemented, is always a safe bet, if it achieves the same thing :)

jl4-lsp/src/LSP/L4/Viz/VizExpr.hs
```diff
@@ -133,7 +135,11 @@ instance HasCodec RenderAsLadderInfo where
     named "RenderAsLadderInfo" $
       object "RenderAsLadderInfo" $
         MkRenderAsLadderInfo
-          <$> requiredField' "funDecl" .= view #funDecl
+          <$> requiredField' "verTextDocId" .= view #verTextDocId
+          <*> requiredField' "funDecl"      .= view #funDecl
+
+instance HasCodec LSP.VersionedTextDocumentIdentifier where
+  codec = codecViaAeson "VersionedTextDocumentIdentifier"
```
MangoIV, line 142: can you inline this instances into the instances you need it in? then we could skip the orphan ~ that's also nicer in the sense that we don't *use* this codec for ser/de of VersionedTextDocumentIdentifier?

MangoIV (approved):
I like this as is, the remaining comments are just nits.

jl4-lsp/src/LSP/L4/Viz/VizExpr.hs
ym-han: I looked briefly into trying to inline the instance, but I think it's fine to stick with this for the time being (The autodocodec piggybacks on the aeson instance, so I don't think there's a real risk of doing some kind of unorthodox serializing/deserializing no matter what.). The orphan instance is not great, but that will be fixed if we switch away from autocodec back to aeson in the future.

PR #400, opened 2025-05-06
Implement floating point literals

### LINKED ISSUE #343: number parsing using decimals is broken
Image

Image

fendor:
We change the 'NUMBER' type to work with rational numbers instead of only integers.
This allows us to use floating point literals, such as `2.0`, and `2.3`, and we still support integer literals, e.g. `1`, `2`.

Since 'NUMBER' is now backed by `Rational`, we cannot represent irrational numbers accurately, but at least to some approximation. The advantage of `Rational` over `Double` is that we do not suffer from IEEE decisions, such as `1.0 / 0.0 = NaN` (or `NaN` in general) and the usual floating point inaccuracies such as `0.1 + 0.2 = 0.3` is indeed `TRUE` in L4.

We introduce the following functions for working with 'NUMBER'

* '`IS INTEGER'`: returns 'TRUE' iff the number is an integer
* 'ROUND': rounds to the closest whole integer
* 'FLOOR': truncates to the next smallest integer
* 'CEILING': increases to the next largest integer

These functions are all built-in into the typechecker and evaluator.

Many other useful utility can be added in user-land or added to the evaluator on a by-need basis.

Implementation wise, most of the patch is rather straight forward, we introduce a new TokenType 'TFloatLit' which parses numbers such as '2.0' and '1.23456789'.
All 'NUMBER's are now 'Rational', so 'TIntLit' is also converted to 'Rational' during parsing, which is a lossless operation.

In the typechecker, we introduce the new built-in functions, which is mostly boilerplate.

Similarly in the evaluator, we provide the implementations for the built-in functions.

Bug fixes included in this patch that are not 100% related to the feature itself:

* Division by zero throws an exception instead of halting the language server
* Module by zero throws an exception instead of halting the langauge server

Closes #343

### DIFF
--- a/jl4-core/src/L4/Evaluate.hs
+++ b/jl4-core/src/L4/Evaluate.hs
@@ -18,11 +18,13 @@ import qualified Base.Map as Map
 import L4.Annotation
 import L4.Evaluate.Operators
 import L4.Evaluate.Value
+import L4.Evaluate.ValueLazy (UnaryBuiltinFun(..))
 import L4.Parser.SrcSpan (SrcRange)
 import L4.Print
 import L4.Syntax
 import qualified L4.TypeCheck as TypeCheck
 import L4.Utils.RevList
+import L4.Utils.Ratio
 
 import Data.Either
 
@@ -115,6 +117,8 @@ data EvalException =
   | EqualityOnUnsupportedType
   | NonExhaustivePatterns Value -- we could try to warn statically
   | StackOverflow
+  | DivisionByZero
+  | NotAnInteger Rational
   | Stuck
     (Expr Resolved)
     -- ^ the expression we were evaluating when getting stuck
@@ -148,6 +152,12 @@ prettyEvalException = \case
     [ "Stack overflow: "
     , "Recursion depth of " <> Text.show maximumStackSize
     , "exceeded." ]
+  DivisionByZero ->
+    [ "Division by zero"
+    ]
+  NotAnInteger num ->
+    [ "Expected an Integer but got the fractional number: " ]
+    <> [ prettyRatio num ]
   Stuck expr r ->
     [ "Expression stuck while evaluating:" ]
     <> prepare expr
@@ -254,6 +264,10 @@ initialEnvironment =
     [ (TypeCheck.falseUnique, falseVal)
     , (TypeCheck.trueUnique,  trueVal)
     , (TypeCheck.emptyUnique, ValList [])
+    , (TypeCheck.isIntegerUnique, ValUnaryBuiltinFun UnaryIsInteger)
+    , (TypeCheck.roundUnique, ValUnaryBuiltinFun UnaryRound)
+    , (TypeCheck.ceilingUnique, ValUnaryBuiltinFun UnaryCeiling)
+    , (TypeCheck.floorUnique, ValUnaryBuiltinFun UnaryFloor)
     ]
 
 evalModule :: Module Resolved -> Eval ()
@@ -484,6 +498,9 @@ backwardExpr !ss stack0@(App1 n vals [] env stack) val = do
       popFrame val
       env'' <- matchGivens givens (reverse (val : vals)) stack0
       pushExprFrame (Map.union env'' env') (ss - 1) stack e
+    Just (ValUnaryBuiltinFun builtin) -> do
+      popFrame val
+      runBuiltin stack0 (reverse (val : vals)) builtin
     Just (ValUnappliedConstructor r) -> do
       popFrame val
       backwardExpr (ss - 1) stack (ValConstructor r (reverse (val : vals)))
@@ -575,11 +592,18 @@ exception exc _stack = do
   throwError exc
 
 runBinOp :: BinOp -> Value -> Value -> Stack -> Eval Value
-runBinOp BinOpPlus   (ValNumber num1) (ValNumber num2) _stack = pure $ ValNumber (num1 + num2)
-runBinOp BinOpMinus  (ValNumber num1) (ValNumber num2) _stack = pure $ ValNumber (num1 - num2)
-runBinOp BinOpTimes  (ValNumber num1) (ValNumber num2) _stack = pure $ ValNumber (num1 * num2)
-runBinOp BinOpDividedBy (ValNumber num1) (ValNumber num2) _stack = pure $ ValNumber (num1 `div` num2)
-runBinOp BinOpModulo    (ValNumber num1) (ValNumber num2) _stack = pure $ ValNumber (num1 `mod` num2)
+runBinOp BinOpPlus      (ValNumber num1) (ValNumber num2) _stack = pure $ ValNumber (num1 + num2)
+runBinOp BinOpMinus     (ValNumber num1) (ValNumber num2) _stack = pure $ ValNumber (num1 - num2)
+runBinOp BinOpTimes     (ValNumber num1) (ValNumber num2) _stack = pure $ ValNumber (num1 * num2)
+runBinOp BinOpDividedBy (ValNumber num1) (ValNumber num2)  stack
+  | num2 /= 0 = pure $ ValNumber (num1 / num2)
+  | otherwise = exception DivisionByZero stack
+runBinOp BinOpModulo    (ValNumber num1) (ValNumber num2) stack  = do
+  n1 <- expectInteger stack num1
+  n2 <- expectInteger stack num2
+  if n2 /= 0
+    then pure $ ValNumber (toRational $ n1 `mod` n2)
+    else exception DivisionByZero stack
 runBinOp BinOpCons   val1             (ValList val2)   _stack = pure $ ValList (val1 : val2)
 runBinOp BinOpEquals val1             val2             stack  = runBinOpEquals val1 val2 stack
 runBinOp BinOpLeq    (ValNumber num1) (ValNumber num2) _stack = pure $ valBool (num1 <= num2)
@@ -634,6 +658,40 @@ valBool True  = trueVal
 runLit :: Lit -> Eval Value
 runLit (NumericLit _ann num) = pure (ValNumber num)
 runLit (StringLit _ann str)  = pure (ValString str)
+
+runBuiltin :: Stack -> [Value] -> UnaryBuiltinFun -> Eval Value
+runBuiltin s vals op = do
+  val :: Rational <- expect1Number s vals
+  pure case op of
+    UnaryIsInteger -> valBool $ isJust $ isInteger val
+    UnaryRound -> valInt $ round val
+    UnaryCeiling -> valInt $ ceiling val
+    UnaryFloor -> valInt $ floor val
+  where
+    valInt :: Integer -> Value
+    valInt = ValNumber . toRational
+
+expect1 :: Stack -> [a] -> Eval a
+expect1 s = \case
+  [x] -> pure x
+  xs ->
+    exception (RuntimeTypeError $ "Expected 1 argument but got " <> Text.show (length xs)) s
+
+expectNumber :: Stack -> Value -> Eval Rational
+expectNumber s = \case
+  ValNumber f -> pure f
+  _ -> exception (RuntimeTypeError "Expected number.") s
+
+expect1Number :: Stack -> [Value] -> Eval Rational
+expect1Number s vs = do
+  v <- expect1 s vs
+  expectNumber s v
+
+expectInteger :: Stack -> Rational -> Eval Integer
+expectInteger stack n = do
+  case isInteger n of
+    Nothing -> exception (NotAnInteger n) stack
+    Just i -> pure i
 
 lookupTerm :: Environment -> Resolved -> Maybe Value
 lookupTerm env r =
--- a/jl4-core/src/L4/Evaluate/ValueLazy.hs
+++ b/jl4-core/src/L4/Evaluate/ValueLazy.hs
@@ -38,12 +38,13 @@ data NF = MkNF (Value NF) | ToDeep
   deriving anyclass NFData
 
 data Value a =
-    ValNumber Int -- for now
+    ValNumber Rational
   | ValString Text
   | ValNil
   | ValCons a a
   | ValClosure (GivenSig Resolved) (Expr Resolved) Environment
   | ValObligation Environment MaybeEvaluated MaybeEvaluated (Maybe (Expr Resolved)) (Expr Resolved)
+  | ValUnaryBuiltinFun UnaryBuiltinFun
   | ValUnappliedConstructor Resolved
   | ValConstructor Resolved [a]
   | ValAssumed Resolved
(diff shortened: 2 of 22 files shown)


jl4-core/src/L4/Evaluate.hs
```diff
@@ -572,11 +588,21 @@ exception exc _stack = do
...
+  | otherwise = exception DivisionByZero stack
+runBinOp BinOpModulo    (ValNumber num1) (ValNumber num2) stack  =
+  case isInteger num1 of
+    Nothing -> exception (NotAnInteger num1) stack
+    Just n1 -> case isInteger num2 of
+      Nothing -> exception (NotAnInteger num2) stack
+      Just n2 ->
+        if n2 /= 0
+          then pure $ ValNumber (toRational $ n1 `mod` n2)
+          else exception DivisionByZero stack
```
fendor, line 605: This could likely be cleaned up with dedicated Eval helpers such as `expectNonZeroInteger` and `expectInteger`. Does that sounds like an improvement?

jl4-core/src/L4/Utils/Ratio.hs
```diff
@@ -0,0 +1,17 @@
...
+
+import Data.Ratio (denominator, numerator)
+import Base
+import qualified Base.Text as Text
+
+isInteger :: Rational -> Maybe Integer
+isInteger r =
+  if denominator r == 1
+    then Just $ numerator r
+    else Nothing
```
MangoIV, line 11: it's kinda interesting, that this isn't in `base`

jl4-decision-service/src/Backend/Jl4.hs
```diff
@@ -92,12 +93,16 @@ literalToExpr = \case
 
 valueToFnLiteral :: (Monad m) => Eval.Value -> ExceptT EvaluatorError m FnLiteral
 valueToFnLiteral = \case
-  Eval.ValNumber i -> pure $ FnLitInt $ fromIntegral i
+  Eval.ValNumber i ->
+    if denominator i == 1
+      then pure $ FnLitInt $ numerator i
+      else pure $ FnLitDouble $ fromRational i
```
MangoIV, line 99: why not use your `isInteger` function here?

jl4-decision-service/src/Backend/Jl4.hs
```diff
@@ -92,12 +93,16 @@ literalToExpr = \case
...
   Eval.ValString t -> pure $ FnLitString t
   Eval.ValList vals -> do
     lits <- traverse valueToFnLiteral vals
     pure $ FnArray lits
   Eval.ValClosure {} -> throwE $ InterpreterError "#EVAL produced function closure."
+  Eval.ValUnaryBuiltinFun {} -> throwE $ InterpreterError "#EVAL produced function closure."
```
MangoIV, line 105: I think it would be good if we could distinguish the error messages, so maybe 
```suggestion
  Eval.ValUnaryBuiltinFun {} -> throwE $ InterpreterError "#EVAL produced builtin function closure."
```

jl4-decision-service/src/Backend/Jl4.hs
```diff
@@ -182,9 +187,9 @@ mkBoolean b =
     True -> l4True
     False -> l4False
 
-mkNumericLit :: Int -> Lit
+mkNumericLit :: Real a => a -> Lit
 mkNumericLit =
-  NumericLit emptyAnno
+  NumericLit emptyAnno . toRational
```
MangoIV, line 192: why this typeclass overloading? Doesn't that mean that if we come from e.g. `Double`, we already lost the precision we opted into `Rational` for?

jl4-core/src/L4/Parser.hs
```diff
@@ -1009,11 +1009,17 @@ list = do
...
   attachAnno $
     NumericLit emptyAnno
-      <$> annoEpa (spacedToken (fmap snd <$> preview #_TIntLit) "Numeric Literal")
+      <$> annoEpa (spacedToken (fmap (toRational . snd) <$> preview #_TIntLit) "Numeric Literal")
+
+floatLit :: Parser Lit
+floatLit =
+  attachAnno $
+    NumericLit emptyAnno
+      <$> annoEpa (spacedToken (fmap snd <$> preview #_TRationalLit) "Float Literal")
```
MangoIV, line 1022: The name is a bit misleading, it's not a floating point number.

jl4-core/src/L4/Lexer.hs
```diff
@@ -257,44 +259,73 @@ directives =
...
+dotDecimal_ c' = do
+  void (satisfy (== '.'))
+  let mkNum = foldl' step (c', 0) . chunkToTokens (Proxy :: Proxy Text)
+      step (a, e') c =
+          ( (a * 10 + fromIntegral (digitToInt c))
+          , (e' - 1)
+          )
+  (\ s -> let (c, e) = mkNum s in (c, e, s)) <$> takeWhile1P (Just "digit") isDigit
+{-# INLINE dotDecimal_ #-}
+
```
MangoIV, line 304: is there a reason why we cannot use [one of the combinators provided by megaparsec](https://hackage.haskell.org/package/megaparsec-9.7.0/docs/Text-Megaparsec-Char-Lexer.html#v:scientific)?

jl4-core/src/L4/EvaluateLazy.hs
```diff
@@ -536,31 +552,53 @@ patternMatchFailure = withPoppedFrame $ \ case
...
+    backward $ valBool $ isJust $ isInteger val
+  UnaryRound -> do
+    val <- expectNumber es
+    backward $ ValNumber $ toRational $ round @Rational @Integer val
+  UnaryCeiling -> do
+    val <- expectNumber es
+    backward $ ValNumber $ toRational $ ceiling @Rational @Integer val
+  UnaryFloor -> do
+    val <- expectNumber es
+    backward $ ValNumber $ toRational $ floor @Rational @Integer val
```
MangoIV, line 571: This seems unnecessarily repetitive, maybe: 
```suggestion
runBuiltin es op = do 
  val :: Rational <- expectNumber es 
  backward case op of 
    UnaryIsInteger -> valBool $ isJust $ isInteger val
    UnaryRound -> valInt $ round val
    UnaryCeiling -> valInt $ ceiling val
    UnaryFloor -> valInt $ floor val
  where 
    valInt :: Integer -> WHNF
    valInt = ValNumber . toRational 
```

jl4-core/src/L4/Evaluate.hs
```diff
@@ -632,6 +655,43 @@ runLit :: Lit -> Eval Value
...
+expect1Number :: Stack -> [Value] -> Eval Rational
+expect1Number s vs = do
+  v <- expect1 s vs
+  expectNumber s v
+
+expectInteger :: Stack -> Rational -> Eval Integer
+expectInteger stack n = do
+  case isInteger n of
+    Nothing -> exception (NotAnInteger n) stack
+    Just i -> pure i
```
MangoIV, line 693: it would be good if the error message included which operation the error stemmed from, in `expectNumber`, it would be good, if the error message would include the "but got" part of the sentence.

jl4-core/src/L4/Evaluate/ValueLazy.hs
```diff
@@ -38,17 +38,25 @@ data NF = MkNF (Value NF) | ToDeep
...
   | ValEnvironment Environment
   deriving stock Show
 
+data UnaryBuiltinFun
+  = UnaryIsInteger
+  | UnaryRound
+  | UnaryCeiling
+  | UnaryFloor
+  deriving stock (Show)
+
```
MangoIV, line 59: can we not reuse this datatype across lazy and strict evaluators? Perhaps not necessary since the strict one is being deprecated anyway, but it would still declutter things. Perhaps the strict value module could depend on the lazy value module.

jl4-core/src/L4/Evaluate.hs
```diff
@@ -632,6 +655,43 @@ runLit :: Lit -> Eval Value
 runLit (NumericLit _ann num) = pure (ValNumber num)
 runLit (StringLit _ann str)  = pure (ValString str)
 
+runBuiltin :: Stack -> [Value] -> UnaryBuiltinFun -> Eval Value
```
MangoIV, line 658: same problem as in EvaluateLazy. I don't think all this repetition is good. 

I'm not too fond of DRY but if it's that much code it's almost impossible to keep them in sync.

jl4-core/src/L4/Parser.hs
```diff
@@ -1009,11 +1009,17 @@ list = do
...
-numericLit :: Parser Lit
-numericLit =
+intLit :: Parser Lit
+intLit =
   attachAnno $
     NumericLit emptyAnno
-      <$> annoEpa (spacedToken (fmap snd <$> preview #_TIntLit) "Numeric Literal")
+      <$> annoEpa (spacedToken (fmap (toRational . snd) <$> preview #_TIntLit) "Numeric Literal")
+
+floatLit :: Parser Lit
```
MangoIV, line 1018: the naming is a bit off. this is not a floating point number.

jl4-core/src/L4/TypeCheck.hs
```diff
@@ -963,13 +963,21 @@ inferExpr' g =
...
-      checkBinOp number number number "TIMES" Times ann e1 e2
+      choose
+        [ checkBinOp number number number "TIMES" Times ann e1 e2
+        ]
     DividedBy ann e1 e2 ->
-      checkBinOp number number number "DIVIDED BY" DividedBy ann e1 e2
+      choose
+        [ checkBinOp number number number "DIVIDED BY" DividedBy ann e1 e2
+        ]
     Modulo ann e1 e2 ->
```
MangoIV, line 981: leftover?

jl4-core/src/L4/TypeCheck/Environment.hs
```diff
@@ -36,6 +36,12 @@ builtinUri = toNormalizedUri (Uri "jl4:builtin")
 -- 31  TRUE
 -- 32  EMPTY
 -- 40  A (type variable in EMPTY)
+-- Number conversion
+-- 100 IS_INTEGER
+-- 103 ROUND
+-- 104 CEILING
+-- 105 FLOOR
```
MangoIV, line 43: if you rebase on main, this can be improved :)

jl4/examples/ok/tests/float.nlg.golden
```diff
@@ -0,0 +1,26 @@
...
+the sum of `ROUND` with `pi` and `FLOOR` with `pi`
+the sum of `pi` and the product of `ROUND` with `pi` and `pi`
+the product of `FLOOR` with `pi` and `FLOOR` with `pi`
+`CEILING` with `pi`
+`IS_INTEGER`
+`IS_INTEGER` with `pi`
+`IS_INTEGER` with 2
+`IS_INTEGER` with 2
+`IS_INTEGER` with 2.00000001
+consider the case distinctions of `IS_INTEGER` with `pi` :  when `TRUE` has  then `pi`. when `FALSE` has  then the sum of 0.1 and 0.2
```
MangoIV, line 26: i guess this would be an oversight for nlg. it seems as though this always generates a `has` for constructors, even though `TRUE` doesn't have any fields.

MangoIV (approved):
Looks good! Just a couple of adjustments I would make to reduce our maintenance burden and make the code easier to read!

jl4-core/src/L4/Evaluate.hs
```diff
@@ -632,6 +658,32 @@ runLit :: Lit -> Eval Value
...
+expect1Value :: Stack -> [Value] -> Eval Value
+expect1Value s = \case
+  [x] -> pure x
+  xs ->
+    exception (RuntimeTypeError $ "Expected 1 argument but got " <> Text.show (length xs)) s
+
+expectNumber :: Stack -> Value -> Eval Rational
+expectNumber s = \case
+  ValNumber f -> pure f
+  _ -> exception (RuntimeTypeError "Expected rational number.") s
```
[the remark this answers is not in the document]
kosmikus: I would say `number` here, not `rational number`.

jl4-core/src/L4/Evaluate.hs
```diff
@@ -148,6 +151,12 @@ prettyEvalException = \case
     [ "Stack overflow: "
     , "Recursion depth of " <> Text.show maximumStackSize
     , "exceeded." ]
+  DivisionByZero ->
+    [ "Division by zero"
+    ]
+  NotAnInteger num ->
+    [ "Expected an Integer but got the floating point number: " ]
```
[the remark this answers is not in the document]
kosmikus: I would say `fractional number` here.

jl4-core/src/L4/Print.hs
```diff
@@ -302,7 +303,7 @@ instance LayoutPrinter a => LayoutPrinter (LocalDecl a) where
 
 instance LayoutPrinter Lit where
   printWithLayout = \case
-    NumericLit _ t -> pretty t
+    NumericLit _ t -> pretty (fromRational @Double t)
```
kosmikus, line 306: Probably use `prettyRatio`.

jl4-core/src/L4/Utils/Ratio.hs
```diff
@@ -0,0 +1,17 @@
...
+
+prettyRatio :: Rational -> Text
+prettyRatio r = case isInteger r of
+  Nothing -> Text.show $ fromRational @Double r
```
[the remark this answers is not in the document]
fendor: How do we want to display `1 % 3`? I think the most intuitive is the lossy way ...

jl4-core/src/L4/Parser.hs
fendor: Is it not? `1.1234567` is a floating point number, afaict.

jl4-core/src/L4/Utils/Ratio.hs
fendor: Yeah, I was also very surprised

jl4-core/src/L4/Parser.hs
MangoIV: it's just a decimal number, from the user's point of view. a floating point number is when you can "move" the number via the exponent, e.g. IEEE754 floating point numbers

jl4-core/src/L4/Utils/Ratio.hs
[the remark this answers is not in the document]
kosmikus: Sure, we'll have to truncate. But ideally, we could display it as `0.33333...` or something alike.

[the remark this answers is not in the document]
kosmikus: My primary objection to displaying via `Double` is not even rounding artifacts but over-eagerness to use scientific notation:
```
ghci> fromRational @Double (1 / 100)
1.0e-2
```

jl4-core/src/L4/Lexer.hs
fendor: Yes, we lose the original characters

jl4-core/src/L4/Evaluate.hs
fendor: This is an internal error, the typechecker should already catch the case of `expect1Number`.

jl4-core/src/L4/Evaluate/ValueLazy.hs
```diff
@@ -38,23 +38,31 @@ data NF = MkNF (Value NF) | ToDeep
   deriving anyclass NFData
 
 data Value a =
-    ValNumber Int -- for now
+    ValNumber Rational -- for now
```
MangoIV, line 41:
```suggestion
    ValNumber Rational
```

absolute nit

jl4-core/src/L4/Lexer.hs
```diff
@@ -266,44 +268,73 @@ directives =
...
-  <|> TPOpen          <$  char '('
-  <|> TPClose         <$  char ')'
-  <|> TCOpen          <$  char '{'
-  <|> TCClose         <$  char '}'
-  <|> TParagraph      <$  char '§'
-  <|> TComma          <$  char ','
-  <|> TSemicolon      <$  char ';'
-  <|> TDot            <$  char '.'
-  <|> TCopy Nothing   <$  char '^'
+      uncurry TRationalLit <$> try rationalLiteral
```
MangoIV, line 316: super nit: indentation is off

MangoIV (approved):
YES!

PR #403, opened 2025-05-07
refactor the evaluator to use an explicit monad and configurations for mutual recursion

MangoIV:
- introduce a new monad `Machine` that only contains the instructions of the machine 
- interpret  this Monad into `Eval` to run the machine 
- make each of the forward and backwards Machine instruction sequences return a `Config` that can the be use to interpret the respecting functions with, to make the mutual recursion explicit.

### DIFF
--- a/jl4-core/jl4-core.cabal
+++ b/jl4-core/jl4-core.cabal
@@ -73,6 +73,7 @@ library
     L4.Evaluate.Value
     L4.Evaluate.ValueLazy
     L4.EvaluateLazy
+    L4.EvaluateLazy.Machine
     L4.EvaluateLazy.ContractFrame
     L4.ExactPrint
     L4.FindDefinition
--- a/jl4-core/src/L4/EvaluateLazy/Machine.hs
+++ b/jl4-core/src/L4/EvaluateLazy/Machine.hs
@@ -1,0 +1,933 @@
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE ViewPatterns #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE GADTs #-}
+module L4.EvaluateLazy.Machine
+( Frame
+, EvalException (..)
+, UserEvalException (..)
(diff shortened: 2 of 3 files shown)


jl4-core/jl4-core.cabal
```diff
@@ -45,6 +45,8 @@ library
     lsp-types,
     megaparsec,
     mtl,
+    free,
+    kan-extensions,
```
MangoIV, line 49: remove

jl4-core/src/L4/EvaluateLazy/Machine.hs
```diff
@@ -0,0 +1,933 @@
...
+
+instance Monad Machine where
+  Config a >>= k = k a
+  Exception e >>= _k = Exception e
+  WithPoppedFrame k' >>= k = WithPoppedFrame (k' >=> k)
+  PushFrame a >>= k = Bind (PushFrame a) (\_ -> k ())
+  Allocate' a >>= k = Bind (Allocate' a) k
+  PokeThunk rf k' >>= k = Bind (PokeThunk rf k') k
+  NewUnique >>= k = Bind NewUnique k
+  Bind m k' >>= k = Bind m (k' >=> k)
```
kosmikus, line 113: Replace by `(>>=) = Bind`.

jl4-core/src/L4/EvaluateLazy/Machine.hs
```diff
@@ -0,0 +1,933 @@
...
+    let
+      pairs = zip rs1 rs2
+    in
+      case pairs of
+        [] -> backward $ ValBool True
+        ((r1, r2) : rss) -> do
+          PushFrame (EqConstructor1 r2 rss)
+          EvalRef r1
+  | otherwise                                           = Backward $ ValBool False
+runBinOpEquals _                       _                = UserException EqualityOnUnsupportedType
```
kosmikus, line 534: Add a comment / TODO that we may want to at least think about whether we can support equality on contracts.

kosmikus (approved):

PR #412, opened 2025-05-08
[chore] do not use self hosted but github runners

MangoIV:


MangoIV:
Didn't this happen because this repo was private? I think for open repos the plans are much more liberal

ym-han:
> Didn't this happen because this repo was private? I think for open repos the plans are much more liberal

Oh I see, it does say 'private repo usage'. OK this makes sense!

MangoIV:
i dunno why it's failing now though, it always seems so arbitrary what does and does not work where, I don't know why we have containers if it is like this...

MangoIV:
closing this for now

ym-han:
> i dunno why it's failing now though, it always seems so arbitrary what does and does not work where, I don't know why we have containers if it is like this...

Sorry it's so arbitrary btw! that must be really frustrating

PR #417, opened 2025-05-09
Ladder backend: Finish implementing the eval app handler

ym-han:
## Key things that this PR does

* eval app handler: Return LSPErrorCodes_ContentModified if server's version does not match version in eval app request from client
* Flesh out the eval app handler on the backend.
  * Prep an 'app expr maker' when the Decide is first visualized, to be used when the client sends an eval App request.
  * When handling the request, we basically extend the module with an eval directive and return the result of that
  * Augment the eval directive with whatever local decls are in scope, to handle the sort of local decl examples below
* Add and expose `execEvalExprInContextOfModule` from EvaluateLazy. This is in the vein of what @kosmikus recommended, except that I filter out any existing directives instead of making a module with an import (because making the import requires a Resolved)
* Add a `moduleTopDecls :: Lens' (Module n) [TopDecl n]` lens to Syntax.hs
* EvalLazy: Make `range` optional in `EvalDirectiveResult` and `EvalDirective` records

Unrelated changes
* Export Optics.Core's (&), (.~) from Base
* Standardize on `vid` instead of `uid` convention in Ladder backend

Note that the current UI is TBD. Will be standardizing some of the UI stuff in follow-up frontend-focused PRs.

I would have tried to make this more of a Rule as @MangoIV had suggested, but I had already implemented most of this by the time the suggestion was raised. And in any case, the way I'm currently doing it already uses rules like that for getting eval deps (i.e.,it does exploit at least some opportunities for caching computation).

## Examples and limitations

Examples that work the way one might expect:
```
{--------------------------------------------------------
-- function is a top decl
--------------------------------------------------------}

GIVEN p IS A BOOLEAN
      q IS A BOOLEAN
GIVETH A BOOLEAN
DECIDE toplvlMyOr IS p OR q

GIVEN arg1 IS BOOLEAN
      arg2 IS BOOLEAN
      arg3 IS BOOLEAN
DECIDE result1 IS
  arg1 AND toplvlMyOr arg2 arg3


{--------------------------------------------------------
-- function is locally defined
--------------------------------------------------------}
GIVEN arg1 IS BOOLEAN
      arg2 IS BOOLEAN
      arg3 IS BOOLEAN
DECIDE result2 IS
  arg1 AND myOr_ arg2 arg3 -- does shadowing with a local def not work?
    WHERE
      myOr_ MEANS GIVEN x, y IS A BOOLEAN YIELD (x OR y)
```

Examples that don't (and probably require further discussion re whether we want to support this sort of thing)

```

{--------------------------------------------------------
Simple edge case involving a locally defined boolean literal
-- Intuitively, one shouldn't be able to change the value of `locallyDefinedArg` in the visualizer

(Yes, there's a similar issue even without local defs, when we visualize boolean literals, but the plan is to add boolean lits and make it impossible to 're-assign' the values of boolean literals.
But just adding boolean literals to the VizExpr/IRExpr AST wouldn't be enough
to get what I think is the intuitive behavior in *this* case.)
--------------------------------------------------------}
GIVEN arg1 IS BOOLEAN
DECIDE resultLocalDefBoolLit IS
  toplvlMyOr arg1 locallyDefinedArg
    WHERE
      locallyDefinedArg MEANS FALSE


{--------------------------------------------------------
Edge case 2: locally defined boolean var that
             is derivative on other boolean vars

Notice that `arg1Andarg2` does not change 
even after the values of arg1 and arg2 have been assigned
---------------------------------------------------------}
GIVEN arg1, arg2 IS BOOLEAN
DECIDE resultDerivedLocalDef IS
  arg2 OR toplvlMyOr arg1 arg1Andarg2
    WHERE
      arg1Andarg2 MEANS arg1 AND arg2 
```

### DIFF
--- a/jl4-core/src/L4/EvaluateLazy.hs
+++ b/jl4-core/src/L4/EvaluateLazy.hs
@@ -166,7 +166,7 @@ nfDirective (MkEvalDirective r expr env) = do
 
 data EvalDirectiveResult =
   MkEvalDirectiveResult
-    { range  :: !SrcRange -- ^ of the (L)EVAL / CONTRACT directive
+    { range  :: !(Maybe SrcRange) -- ^ of the (L)EVAL / CONTRACT directive
     , result :: Either EvalException NF
     }
   deriving stock (Generic, Show)
--- a/jl4-core/src/L4/EvaluateLazy.hs
+++ b/jl4-core/src/L4/EvaluateLazy.hs
@@ -166,7 +166,7 @@ nfDirective (MkEvalDirective r expr env) = do
 
 data EvalDirectiveResult =
   MkEvalDirectiveResult
-    { range  :: !(Maybe SrcRange) -- ^ of the (L)EVAL / CONTRACT directive
+    { range  :: Maybe SrcRange -- ^ of the (L)EVAL / CONTRACT directive
     , result :: Either EvalException NF
     }
   deriving stock (Generic, Show)
--- a/jl4-core/src/L4/EvaluateLazy.hs
+++ b/jl4-core/src/L4/EvaluateLazy.hs
@@ -2,6 +2,7 @@
 module L4.EvaluateLazy
 ( EvalDirectiveResult (..)
 , execEvalModuleWithEnv
+, execEvalExprInContextOfModule
 , prettyEvalException
 )
 where
@@ -13,6 +14,7 @@ import L4.EvaluateLazy.Machine
 import Control.Concurrent
 import L4.Evaluate.ValueLazy
 import L4.Parser.SrcSpan (SrcRange)
+import L4.Annotation
 import L4.Syntax
 
 -----------------------------------------------------------------------------
@@ -242,3 +244,31 @@ evalModuleAndDirectives env m = do
   -- Depending on future export semantics, this may have to change.
   pure (env', results)
 
+
+{- | Evaluate an expression in the context of a module and initial environment.
+
+Didn't try to cache even more computation with rules, 
+because the current Rule type seems to
+be Uri-focused, and so you'll emd up needing to pretty print and then re-parse.
+Also, it's not clear how much caching can actually be done,
+given that we won't be re-using the result from this.
+ -}
+execEvalExprInContextOfModule :: Expr Resolved -> (Environment, Module Resolved) -> IO (Maybe EvalDirectiveResult)
+execEvalExprInContextOfModule expr (env, m) = do
+  let
+    evalExprDirective =
+      Directive emptyAnno $ LazyEval emptyAnno expr
+    -- Didn't make a new module that imported the context module,
+    -- because making the import requires a Resolved.
+    moduleWithoutDirectives = over moduleTopDecls (filter $ not . isDirective) m
+  (_, res) <- execEvalModuleWithEnv env (evalExprDirective `prependToModule` moduleWithoutDirectives)
+  case res of
+    [result] -> pure (Just result)
+    _        -> pure Nothing
+  where
+    isDirective :: TopDecl Resolved -> Bool
+    isDirective (Directive _ _) = True
+    isDirective _ = False
+
+    prependToModule :: TopDecl Resolved -> Module Resolved -> Module Resolved
+    prependToModule newDecl = over moduleTopDecls (newDecl :)
--- a/jl4-core/src/L4/EvaluateLazy/Machine.hs
+++ b/jl4-core/src/L4/EvaluateLazy/Machine.hs
@@ -898,7 +898,7 @@ emptyEnvironment = Map.empty
 
 data EvalDirective =
   MkEvalDirective
-    { range :: !(Maybe SrcRange) -- ^ of the (L)EVAL directive
+    { range :: Maybe SrcRange -- ^ of the (L)EVAL directive
     , expr  :: !(Expr Resolved) -- ^ expression to evaluate
     , env   :: !Environment -- ^ environment to evaluate the expression in
     }
--- a/jl4-core/src/L4/EvaluateLazy/Machine.hs
+++ b/jl4-core/src/L4/EvaluateLazy/Machine.hs
@@ -18,6 +18,7 @@ module L4.EvaluateLazy.Machine
 , evalRef
 , emptyEnvironment
 , prettyEvalException
+, boolView
 )
 where
 
@@ -596,7 +597,7 @@ valBool False = falseVal
 valBool True  = trueVal
 
 -- | Checks if a value is a Boolean constructor.
-boolView :: WHNF -> Maybe Bool
+boolView :: Value a -> Maybe Bool
 boolView val =
   case val of
     ValConstructor n []
--- a/jl4-core/src/L4/EvaluateLazy/Machine.hs
+++ b/jl4-core/src/L4/EvaluateLazy/Machine.hs
@@ -758,7 +758,7 @@ evalTopDecl _env (Import _ann _import_) =
 
 evalDirective :: Environment -> Directive Resolved -> Machine [EvalDirective]
 evalDirective env (LazyEval ann expr) =
-  pure ((\ r -> MkEvalDirective r expr env) <$> toList (rangeOf ann))
+  pure [MkEvalDirective (rangeOf ann) expr env]
 evalDirective _env (StrictEval _ann _expr) =
   pure []
 evalDirective _env (Check _ann _expr) =
@@ -898,7 +898,7 @@ emptyEnvironment = Map.empty
 
 data EvalDirective =
   MkEvalDirective
-    { range :: !SrcRange -- ^ of the (L)EVAL directive
+    { range :: !(Maybe SrcRange) -- ^ of the (L)EVAL directive
     , expr  :: !(Expr Resolved) -- ^ expression to evaluate
     , env   :: !Environment -- ^ environment to evaluate the expression in
     }
--- a/jl4-lsp/app/LSP/L4/Handlers.hs
+++ b/jl4-lsp/app/LSP/L4/Handlers.hs
@@ -349,27 +349,23 @@ handlers recorder =
 
             mRecentViz <- liftIO $ atomically $ getMostRecentVisualisation ide
             case mRecentViz of
+              Just recentViz
+                | evalParams.verDocId._version == recentViz.vizEnv.verTxtDocId._version ->
+                    -- Haven't implemented the actual call to eval yet; just returning null for now
+                    pure $ Right Aeson.Null
+                | otherwise -> -- Version mismatch
+                    pure $ Left $ TResponseError
+                       { _code = InL LSPErrorCodes_ContentModified
+                       , _message = "Document version mismatch. Visualizer version: " <> Text.pack (show evalParams.verDocId._version) <>
+                                    ", whereas server's version is: " <> Text.pack (show recentViz.vizEnv.verTxtDocId._version)
+                       , _xdata = Nothing
+                       }
               Nothing -> -- impossible
                 pure $ Left $ TResponseError
                   { _code = InR ErrorCodes_InvalidRequest
                   , _message = "No recent visualisation found, when trying to handle " <> methodName <> ". This case should be impossible."
                   , _xdata = Nothing
                   }
-                  
-              Just recentViz -> do
-                let clientVersion = evalParams.verDocId._version
-                    serverVersion = recentViz.vizEnv.verTxtDocId._version
-
-                if clientVersion == serverVersion then
-                  -- Haven't implemented the actual call to eval yet; just returning null for now
-                  pure $ Right Aeson.Null
-                else 
-                  pure $ Left $ TResponseError
-                                  { _code = InL LSPErrorCodes_ContentModified
-                                  , _message = "Document version mismatch. Visualizer version: " <> Text.pack (show clientVersion) <>
-                                              ", whereas server's version is: " <> Text.pack (show serverVersion)
-                                  , _xdata = Nothing
-                                  }
     ]
 
 activeFileDiagnosticsInRange :: ShakeExtras -> NormalizedUri -> Range -> STM [FileDiagnostic]
--- a/jl4-lsp/app/LSP/L4/Handlers.hs
+++ b/jl4-lsp/app/LSP/L4/Handlers.hs
@@ -353,8 +353,9 @@ handlers recorder =
             
             runExceptT $ case (mtcRes, mRecentViz) of
               -- The following two cases should be impossible,
-              -- since the verTxtDocId that the client sends us in the l4/evalApp request
-              -- corresponds to the same one that it got when it was requested to render the VizExpr.
+              -- since the client can only send an l4/evalApp request for an App in a VizExpr that it had antecedently been requested to render.
+              -- (And once the server asks the client to render the VizExpr,
+              -- the server will store the associated RecentViz.)
               (Nothing, _) -> defaultResponseError $ "Failed to get type check for " <> Text.show evalParams.verDocId._uri
               (_, Nothing) -> defaultResponseError $ "No recent visualisation found, when trying to handle " <> methodName <> ". This case should be impossible."
               
@@ -362,11 +363,13 @@ handlers recorder =
               -- and use oneshotL4ActionAndErrors on that new file
               
               {- We require that the client's verTxtDocId matches the server's.
-                 Note, again, that the client will have already received 
-                 the verTxtDocId in the original 'please render this VizExpr' request -}
-              (Just tcRes, Just recentViz)
-                | let vizConfig = Ladder.getVizConfig . (.vizState) $ recentViz,
-                  evalParams.verDocId == vizConfig.verTxtDocId -> do
+                 Note that the verTxtDocId that the client sends us in the l4/evalApp request
+                 corresponds to the one that it got when it was requested to render the VizExpr.
+              -}
+              (Just tcRes, Just recentViz) ->
+                let vizConfig = Ladder.getVizConfig . (.vizState) $ recentViz
+                in if evalParams.verDocId == vizConfig.verTxtDocId 
+                  then do
                     let nuri = vizConfig.moduleUri
                     mEvalDeps <- liftIO $ runAction "l4/evalApp" ide $ use (AttachCallStack [nuri] GetLazyEvaluationDependencies) nuri
                     case mEvalDeps of
@@ -380,11 +383,13 @@ handlers recorder =
                         logWith recorder Debug $ LogReceivedCustomRequest evalParams.verDocId._uri 
                           ("Eval result: " <> Text.pack (show result))
                         pure result
-                | otherwise -> throwError $ TResponseError 
-                    -- TODO: Have the client update accordingly when it gets this error code,
-                    -- if it doesn't alr do so automatically
+                  else
+                    throwError $ TResponseError 
+                      -- TODO: Have the client update accordingly when it gets this error code,
+                      -- if it doesn't alr do so automatically
                     { _code = InL LSPErrorCodes_ContentModified
-                    , _message = "Document version mismatch. Visualizer version: " <> Text.pack (show evalParams.verDocId._version)
+                    , _message = "Document version mismatch. Visualizer version: " <> Text.pack (show evalParams.verDocId._version) <>
+                    ", whereas server's version is: " <> Text.pack (show vizConfig.verTxtDocId._version)
                     , _xdata = Nothing
                     }
     ]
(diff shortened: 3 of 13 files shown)


jl4-lsp/src/LSP/L4/Viz/Ladder.hs
```diff
@@ -30,67 +49,108 @@ isDevMode = False
...
-data VizState =
-  MkVizState
-    { env :: !VizEnv
-    , maxId :: !ID
-    }
-  deriving stock (Show, Generic, Eq)
 
-mkInitialVizState :: VizEnv -> VizState
-mkInitialVizState env =
+data VizState = MkVizState
```
ym-han, line 79: I might be pre-maturely optimizing with the bang patterns --- haven't done any profiling

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
```diff
@@ -30,67 +49,108 @@ isDevMode = False
...
 
 getShouldSimplify :: Viz Bool
 getShouldSimplify = do
-  env <- getVizEnv
-  pure env.shouldSimplify
+  cfg <- getVizCfg
+  pure cfg.shouldSimplify
+
+{- | Assumes that the program is well-scoped and well-typed -}
+prepEvalAppMaker :: V.ID -> Expr Resolved -> Viz ()
```
ym-han, line 127: This function is probably one of the most important parts of the PR

jl4-core/src/L4/EvaluateLazy.hs
```diff
@@ -166,7 +166,7 @@ nfDirective (MkEvalDirective r expr env) = do
 
 data EvalDirectiveResult =
   MkEvalDirectiveResult
-    { range  :: !SrcRange -- ^ of the (L)EVAL / CONTRACT directive
+    { range  :: !(Maybe SrcRange) -- ^ of the (L)EVAL / CONTRACT directive
```
MangoIV, line 169: doesn't need to be strict anymore as it doesn't make the SrcRange strict.

jl4-core/src/L4/EvaluateLazy/Machine.hs
```diff
@@ -898,7 +898,7 @@ emptyEnvironment = Map.empty
 
 data EvalDirective =
   MkEvalDirective
-    { range :: !SrcRange -- ^ of the (L)EVAL directive
+    { range :: !(Maybe SrcRange) -- ^ of the (L)EVAL directive
```
MangoIV, line 901: same here, doesn't need to be strict anymore.

jl4-lsp/app/LSP/L4/Handlers.hs
```diff
@@ -333,19 +333,58 @@ handlers recorder =
...
+          
+          -- eval params valid
+          Aeson.Success evalParams -> do
+            logWith recorder Debug $ LogReceivedCustomRequest evalParams.verDocId._uri methodName
+            
+            mtcRes <- liftIO $ runAction "l4/evalApp" ide $ use TypeCheck $ toNormalizedUri evalParams.verDocId._uri
+            mRecentViz <- liftIO $ atomically $ getMostRecentVisualisation ide
+            
+            runExceptT $ case (mtcRes, mRecentViz) of
+              -- impossible
```
MangoIV, line 355: what does "impossible" mean here?

jl4-lsp/app/LSP/L4/Handlers.hs
```diff
@@ -333,19 +333,58 @@ handlers recorder =
...
+              (Nothing, _) -> defaultResponseError $ "Failed to get type check for " <> Text.show evalParams.verDocId._uri
+              (_, Nothing) -> defaultResponseError $ "No recent visualisation found, when trying to handle " <> methodName <> ". This case should be impossible."
+              
+              -- Another approach would be to make a new virtual file with the existing module + our directive,
+              -- and use oneshotL4ActionAndErrors on that new file
+              
+              {- We require that the client's verTxtDocId matches the server's.
+                 Note that the client will have already received 
+                 the verTxtDocId in the original payload for the 'please render this VizExpr' request -}
+              (Just tcRes, Just recentViz@(Ladder.getVizConfig . (.vizState) -> vizConfig))
```
MangoIV, line 365: please don't use viewpatterns if you don't actually match any patterns

jl4-lsp/app/LSP/L4/Handlers.hs
```diff
@@ -333,19 +333,58 @@ handlers recorder =
...
+          Aeson.Success evalParams -> do
+            logWith recorder Debug $ LogReceivedCustomRequest evalParams.verDocId._uri methodName
+            
+            mtcRes <- liftIO $ runAction "l4/evalApp" ide $ use TypeCheck $ toNormalizedUri evalParams.verDocId._uri
+            mRecentViz <- liftIO $ atomically $ getMostRecentVisualisation ide
+            
+            runExceptT $ case (mtcRes, mRecentViz) of
+              -- impossible
+              (Nothing, _) -> defaultResponseError $ "Failed to get type check for " <> Text.show evalParams.verDocId._uri
+              (_, Nothing) -> defaultResponseError $ "No recent visualisation found, when trying to handle " <> methodName <> ". This case should be impossible."
```
MangoIV, line 357: I assume this case is impossible because the client ensures that it first requests a visualization and only then uses the custom method? In that case the error message or documentation should contain that fact.

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
```diff
@@ -30,67 +49,108 @@ isDevMode = False
...
+  , evalAppMakers :: !(Map Int (V.EvalAppRequestParams -> TopDecl Resolved))
+  -- ^ Map from Unique of V.ID to eval-app-directive maker
+  }
+  deriving stock (Generic)
+
+instance Show VizState where
+  show MkVizState{cfg, maxId, evalAppMakers} =
+    "MkVizState { cfg = " <> show cfg <>
+    ", maxId = " <> show maxId <>
+    ", (keys of) evalAppMakers = " <> show (Map.keys evalAppMakers) <> " }"
```
MangoIV, line 91: why do we need this custom Show instance?

jl4-lsp/src/LSP/L4/Actions.hs
```diff
@@ -96,6 +103,54 @@ gotoDefinition pos m positionMapping = do
...
+  -> Ladder.EvalAppRequestParams
+  -> RecentlyVisualised
+  -> EL.Environment
+  -> ExceptT (TResponseError method) m Aeson.Value
+evalApp tcRes evalParams recentViz evalEnv =
+  case Ladder.lookupEvalAppMaker recentViz.vizState evalParams.appExpr of
+    Nothing -> defaultResponseError "No eval app directive maker found" -- TODO: Improve error codehere
+    Just evalAppMaker -> do
+      let evalAppDirective = evalAppMaker evalParams
+      (_, results) <- liftIO $ EL.execEvalModuleWithEnv evalEnv (evalAppDirective `prependToModule` tcRes.module')
```
MangoIV, line 123: If you don't want to use the rule directly and that's why you decided to use the function here instead, you can pass in a continuation that uses the rule from the callsite of this function and then just use the continuation in place of execEvalModuleWithEnv

jl4-lsp/src/LSP/L4/Actions.hs
```diff
@@ -111,38 +166,50 @@ visualise
   -> ExceptT (TResponseError method) m (Aeson.Value |? Null)
 visualise mtcRes (getRecVis, setRecVis) verTextDocId msrcPos = do
   let uri = verTextDocId._uri
-  mdecide :: Maybe (Decide Resolved, Bool, Substitution) <- case msrcPos of
-    -- the command was issued by the button in vscode or autorefresh
+
+  -- Try to pinpoint a Decide (and VizConfig) based on how the command was issued (autorefresh vs code action/code lens)
```
ym-han, line 170: Added some comments for myself

jl4-lsp/src/LSP/L4/Actions.hs
```diff
@@ -96,6 +103,54 @@ gotoDefinition pos m positionMapping = do
   newRange <- toCurrentRange positionMapping lspRange
   pure (Location (fromNormalizedUri range.moduleUri) newRange)
 
+
+-- ----------------------------------------------------------------------------
+-- Ladder evalApp
+-- ----------------------------------------------------------------------------
+
+evalApp
+  :: MonadIO m
```
MangoIV, line 112: you can scope your `m` here with `forall m.`, then you don't get a new `m` in each decl in the `where` (that should save you the constraints)

jl4-lsp/app/LSP/L4/Handlers.hs
```diff
@@ -333,19 +333,58 @@ handlers recorder =
...
+                        pure result
+                | otherwise -> throwError $ TResponseError 
+                    -- TODO: Have the client update accordingly when it gets this error code,
+                    -- if it doesn't alr do so automatically
+                    { _code = InL LSPErrorCodes_ContentModified
+                    , _message = "Document version mismatch. Visualizer version: " <> Text.pack (show evalParams.verDocId._version) <>
+                                ", whereas server's version is: " <> Text.pack (show vizConfig.verTxtDocId._version)
+                    , _xdata = Nothing
+                    }
     ]
```
ym-han, line 389: Let me know if the error codes seem fine? (I'm quite new to the LSP)

jl4-lsp/src/LSP/L4/Actions.hs
```diff
@@ -96,6 +103,54 @@ gotoDefinition pos m positionMapping = do
...
+      Left err           -> defaultResponseError $ Text.unlines $ EL.prettyEvalException err
+
+    throwExpectBoolResultError :: Monad m => ExceptT (TResponseError method) m a
+    throwExpectBoolResultError = defaultResponseError "Ladder visualizer is expecting a boolean result (and it should be impossible to have got a fn with a non-bool return type in the first place)"
+
+    toUBoolValue :: Monad m => Resolved -> ExceptT (TResponseError method) m Ladder.UBoolValue
+    toUBoolValue resolved = case getUnique resolved of
+      u | u == falseUnique -> pure Ladder.FalseV
+        | u == trueUnique  -> pure Ladder.TrueV
+        | otherwise        -> throwExpectBoolResultError
```
MangoIV, line 146: you can use `boolView` instead of this

jl4-lsp/src/LSP/L4/Actions.hs
```diff
@@ -96,6 +103,54 @@ gotoDefinition pos m positionMapping = do
...
+    prependToModule newDecl (MkModule ann nuri (MkSection sann sresolved maka decls)) =
+      MkModule ann nuri (MkSection sann sresolved maka (newDecl : decls))
+
+    evalResultToLadderEvalAppResult :: Monad m => EL.EvalDirectiveResult -> ExceptT (TResponseError method) m EvalAppResult
+    evalResultToLadderEvalAppResult (EL.MkEvalDirectiveResult _ res) = case res of
+      Right (EL.MkNF val) -> 
+        case val of
+          EL.ValConstructor r [] -> EvalAppResult <$> toUBoolValue r
+          _                      -> throwExpectBoolResultError
+      Right EL.ToDeep    -> throwExpectBoolResultError
```
MangoIV, line 136: I don't think that all of these should have the same error, especially the ToDeep one should not be "impossible", should it?

jl4-lsp/src/LSP/L4/Actions.hs
```diff
@@ -96,6 +103,54 @@ gotoDefinition pos m positionMapping = do
...
+    throwExpectBoolResultError :: Monad m => ExceptT (TResponseError method) m a
+    throwExpectBoolResultError = defaultResponseError "Ladder visualizer is expecting a boolean result (and it should be impossible to have got a fn with a non-bool return type in the first place)"
+
+    toUBoolValue :: Monad m => Resolved -> ExceptT (TResponseError method) m Ladder.UBoolValue
+    toUBoolValue resolved = case getUnique resolved of
+      u | u == falseUnique -> pure Ladder.FalseV
+        | u == trueUnique  -> pure Ladder.TrueV
+        | otherwise        -> throwExpectBoolResultError
+
+    -- | Assumes that the order of the eval results is the same as the order of the eval directives.
```
MangoIV, line 148: have you checked that this assumption upholds on modules with multiple eval directives?

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
```diff
@@ -1,24 +1,43 @@
 {-# LANGUAGE ViewPatterns #-}
-
-module LSP.L4.Viz.Ladder where
+{-# OPTIONS_GHC -Wno-unrecognised-pragmas #-}
+{-# HLINT ignore "Redundant lambda" #-}
+
```
MangoIV, line 4:
```suggestion
```

there are hlint hints all over the place, shouldn't be starting to ignore them on a per file basis now

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
```diff
@@ -30,67 +49,108 @@ isDevMode = False
...
-    }
-  deriving stock (Show, Generic, Eq)
 
-mkInitialVizState :: VizEnv -> VizState
-mkInitialVizState env =
+data VizState = MkVizState
+  { cfg           :: !VizConfig
+  , maxId         :: !ID
+  , evalAppMakers :: !(Map Int (V.EvalAppRequestParams -> TopDecl Resolved))
+  -- ^ Map from Unique of V.ID to eval-app-directive maker
```
MangoIV, line 83: If this is `Int` anyway, might as well use an `IntMap` (doesn't need to be the strict one because forcing a function does nothing and the maps are spine strict anyway)

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
```diff
@@ -30,67 +49,108 @@ isDevMode = False
...
+  show MkVizState{cfg, maxId, evalAppMakers} =
+    "MkVizState { cfg = " <> show cfg <>
+    ", maxId = " <> show maxId <>
+    ", (keys of) evalAppMakers = " <> show (Map.keys evalAppMakers) <> " }"
+
+mkInitialVizState :: VizConfig -> VizState
+mkInitialVizState cfg =
   MkVizState
-    { env = env
+    { cfg = cfg
```
MangoIV, line 96:
```suggestion
    { cfg
```

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
```diff
@@ -264,3 +331,13 @@ isBooleanType ty = do
...
+
+------------------------------------------------------
+-- UBoolValue x L4 True/False conversion helpers
+------------------------------------------------------
+
+toBoolExpr :: V.UBoolValue -> Expr Resolved
+toBoolExpr = \ case
+  V.FalseV   -> App emptyAnno TC.falseRef []
+  V.TrueV    -> App emptyAnno TC.trueRef []
+  V.UnknownV -> error "impossible for now"
```
MangoIV, line 343: then why have this constructor?

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
```diff
@@ -221,19 +289,18 @@ scanOr e = [e]
 defaultUBoolVarValue :: V.UBoolValue
 defaultUBoolVarValue = V.UnknownV
 
-leafFromVizName :: V.Name -> Viz IRExpr
-leafFromVizName vname = do
-  uid <- getFresh
-  pure $ V.UBoolVar uid vname defaultUBoolVarValue
+leafFromVizName :: V.ID -> V.Name -> Viz IRExpr
+leafFromVizName vid vname = do
+  pure $ V.UBoolVar vid vname defaultUBoolVarValue
```
MangoIV, line 294: This function should probably not be effectful, if it doesn't even "conceptually" require to be

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
```diff
@@ -30,67 +49,108 @@ isDevMode = False
...
+withLocalDecls :: [LocalDecl Resolved] -> Viz a -> Viz a
+withLocalDecls newLocalDecls = local (\env -> env { localDecls = newLocalDecls <> env.localDecls })
+
+-- Viz state helpers
+
+lookupEvalAppMaker :: VizState -> V.ID -> Maybe (V.EvalAppRequestParams -> TopDecl Resolved)
+lookupEvalAppMaker vs vid = Map.lookup vid.id vs.evalAppMakers
+
+getVizConfig :: VizState -> VizConfig
+getVizConfig vs = vs.cfg
```
MangoIV, line 153: I don't understand why you use this mix of record dot accessors and functions that are defined in terms of them.

jl4-core/src/L4/EvaluateLazy.hs
```diff
@@ -166,7 +166,7 @@ nfDirective (MkEvalDirective r expr env) = do
```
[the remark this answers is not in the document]
MangoIV: yes, that is okay

jl4-lsp/app/LSP/L4/Handlers.hs
ym-han: Good point thank you. You've pointed this out before, sorry for not having internalized this

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
ym-han: Thank you!! very useful to know

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
ym-han: Nice catch!

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
ym-han: I wanted to hide the implementational details of VizState (e.g. how VizState is related to VizConfig).

I've pushed a commit with a comment explaining this, and removed the export of the internals of VizState.

jl4-core/src/L4/EvaluateLazy.hs
ym-han: Thanks for pointing this out!

jl4-lsp/src/LSP/L4/Actions.hs
ym-han: Nice. I've managed to get the `m` in wide scope / to be the same throughout, but I don't know how to get the right type for `method` (it shouldn't be *any* method). 

```
evalApp
  :: forall m method.
  (MonadIO m)
  => TypeCheckResult
  -> Ladder.EvalAppRequestParams
  -> RecentlyVisualised
  -> EL.Environment
  -> ExceptT (TResponseError method) m Aeson.Value
```
Then again, it also isn't a big deal whether we get the `method` type param right

jl4-lsp/src/LSP/L4/Actions.hs
ym-han: I'm using the execEval... function because I didn't see an easy way to use the `EvaluateLazy` rule on a new module (that consists of the evalAppDirective prepended to the old). But I do use the GetLazyEvaluationDependencies rule in the handler to get the eval environment and then pass that in to the `evalApp` call.

The EvalLazy rule, for reference:
```
  define shakeRecorder $ \EvaluateLazy uri -> do
    res  <- use_ (AttachCallStack [uri] GetLazyEvaluationDependencies) uri
    let results = snd res
    pure (mkSimpleFileDiagnostic uri . evalLazyResultToDiagnostic <$> results, Just results)
  ```

ym-han: I could also make a new rule with the eval-related logic from the `evalApp` function, if you think that's better.

jl4-lsp/src/LSP/L4/Actions.hs
ym-han: I changed it to `defaultResponseError "Evaluation exceeded maximum depth limit"`, but let me know if you think of a better error message!

jl4-lsp/src/LSP/L4/Actions.hs
ym-han: Part of what I'm struggling with here is that it's not clear to me what the underlying motivation / concern is exactly. E.g. is it about caching computation as much as possible (in which case I should probably factor the key logic from evalApp out into a new rule)? Or is it more about not using `execEvalModuleWithEnv` directly? Or is it about having all the rules-related logic in one place?

PR #426, opened 2025-05-14
Infra: Remove the confusingly named root `examples` dir; move stuff elsewhere

ym-han:
Depends on #422  --- please review that first


MangoIV (approved):
very nice, I like this.

PR #428, opened 2025-05-14
Viz: Don't show Viz codelens (by throwing error) if Decide does not return a boolean (#419)

### LINKED ISSUE #419: show visualizer only for DECIDE stanzas
while `DECIDE` and `MEANS` are synonyms, in practice I think the visualizer only needs to be offered for a `DECIDE` stanza.

Because it feels like clutter when it shows up as MEANS.

Image

Thoughts?

ym-han:
Closes #419 

* Remove obsoleted VizError variant
* Throw InvalidDecideMustHaveBoolRetType error if the Decide under consideration does not return a boolean


ym-han:
I'm just going to merge this in, because it's a fairly simple change

PR #434, opened 2025-05-15
Viz: More App node UI

ym-han:
Just some more quick app node UI stuff. This includes the ui that was demo'd at the most recent group meeting.

* Viz: Increase app node border by 1 px, relative to uboolvar node border
* Viz: Use bg-gray-100 background for app nodes to evoke the idea of a fn being a 'black box' (but not using solid black b/c don't want too much contrast between this and a uboolvarnode)
* Refactor: extract ladder border width into a css var

image

### DIFF
--- a/ts-apps/decision-logic-visualizer/src/style.css
+++ b/ts-apps/decision-logic-visualizer/src/style.css
@@ -169,6 +169,8 @@
   Utility SF styles
 *************************************/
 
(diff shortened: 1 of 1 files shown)


johsi-k:
Visualising result1 works as expected

johsi-k (approved):

ym-han:
> Visualising result1 works as expected

Thanks!

PR #436, opened 2025-05-16
Mangoiv/or on contracts

MangoIV:


mengwong:
i would be grateful if this PR could make it into main and then onto the web IDE sometime today, as we have a demo lined up for tomorrow that would look much better with LEST support.

jl4-core/src/L4/Evaluate/ValueLazy.hs
```diff
@@ -68,6 +73,7 @@ data UnaryBuiltinFun
 instance NFData a => NFData (Value a) where
   rnf :: Value a -> ()
   rnf (ValNumber i)               = rnf i
+  rnf (ValROp !_env !_op a b)     = a `deepseq` b `deepseq` ()
```
fendor, line 76: that feels unnecessarily subtle. Let's rather be explicit

jl4-core/src/L4/TypeCheck/Types.hs
```diff
@@ -583,18 +584,28 @@ prune m = do
...
-      procPlain a ((Plain _, _)  : []) = [first (With (MkCheckErrorWithContext InternalAmbiguityError ctx)) a]
-      procPlain _ ((Plain _, _)  : cs) = [last cs]
       procPlain a ((With _ _, _) : cs) = procPlain a cs
+      procPlain a cs
+        | plain = [first (With (MkCheckErrorWithContext InternalAmbiguityError ctx)) a]
+        -- NOTE: in the case of ambiguity errors, the ambiguity error will always occurs as
+        -- the last element in the list
+        | otherwise = [last cs]
+        where
+          plain = allPlain $ map fst cs
```
fendor, line 594: Is this correct?
Wouldn't that mean that '[Plain, Plain, With]' reports 'With', even though it should be ambiguous?

jl4-core/src/L4/TypeCheck/Types.hs
```diff
@@ -583,18 +584,28 @@ prune m = do
...
 
       -- We have a failure, we're still looking for a success, and prefer the last failure
-      procWith a []                     = [a]
+      procWith a []                                     = [a]
```
fendor, line 597: Alignment

jl4-core/src/L4/TypeCheck/Types.hs
```diff
@@ -583,18 +584,28 @@ prune m = do
...
       procWith _ ((Plain a, s')  : cs)  = procPlain (Plain a, s') cs
-      procWith _ ((With e x, s') : cs)  = procWith (With e x, s') cs
+      procWith _ ((With e x, s') : cs)                  = procWith (With e x, s') cs
```
fendor, line 599: Alignment

jl4-core/src/L4/TypeCheck/Types.hs
```diff
@@ -612,8 +623,7 @@ softprune m = do
 
       -- We have a success, we don't want a second one
       procPlain a []                    = [a]
-      procPlain _ ((Plain _, _)  : [])  = candidates
-      procPlain _ ((Plain _, _)  : _cs) = candidates
+      procPlain _ ((Plain _, _)  : _cs)  = candidates
```
fendor, line 626:
```suggestion
      procPlain _ ((Plain _, _)  : _cs) = candidates
```

fendor (approved):
LGTM, couple nitpicks :)

jl4-core/src/L4/TypeCheck/Types.hs
MangoIV: It would report With but then the With would contain an ambiguity error. That's exactly the assumption that we're (ab)using

PR #440, opened 2025-05-19
Viz: Extend client-side RPC with custom `inlineExprs` request and add placeholder UI for it in the Ladder visualizer

ym-han:
This is mostly a Ladder frontend PR; the backend handler will be implemented in the next PR

* Backend: Add `canInline` field to UBoolVar that currently just defaults to False
* Frontend VizExpr/IRExpr: add `canInline` field to UBoolVar in IRExpr/VizExpr
* Extend Ladder client-side RPC to support inlining exprs
* Doc: Explain how to add support for a custom LSP request on the client side
* Frontend: Wire up the viz to client-side RPC. I.e.,  add _placeholder_ UI for `canInline` and call the `inlineExprs` L4Conn method when the uboolvar node is clicked

@johsi-k will be experimenting with and polishing the UI in subsequent PR(s)

Unrelated: 
* Refactor / Fix: Have consumers of the visualizer lib define the LirRootType themselves and pass it into the visualizer via the LadderEnv

PR #446, opened 2025-05-19
Ladder backend: Add a prelim VizHandler monad, and refactor eval app handler to use that. Stub out the inlineExprs handler

ym-han:
Depends on #440 

VizHandler refactor
* Add VizHandler monad and helpers, and IsLadderRequestParams typeclass
* Handlers.hs: Refactor the eval app handler to use VizHandler

Params, payloads types for inlining exprs (+ some clean up of eval app tojson)
* CustomProtocol.hs: Fix EvalAppResult's ToJSon encoding
* CustomProtocol.hs: Add custom method type for inlining exprs
* Stub out the InliningExprs handler in the VizHandler monad

Will implement the logic for whether this or that subexpr's `canInline` is true/false in the next PR, along with the handler for inlining exprs

### DIFF
--- a/jl4-lsp/app/LSP/L4/Handlers.hs
+++ b/jl4-lsp/app/LSP/L4/Handlers.hs
@@ -46,7 +46,7 @@ import qualified LSP.Core.Shake as Shake
 import LSP.Core.Types.Diagnostics
 import LSP.Core.Types.Location
 import qualified LSP.L4.Viz.Ladder as Ladder
-import LSP.L4.Viz.CustomProtocol as Ladder
+import           LSP.L4.Viz.CustomProtocol as Ladder
 import LSP.Logger
 import qualified Language.LSP.Protocol.Lens as J
 import Language.LSP.Protocol.Message
@@ -335,7 +335,7 @@ handlers recorder =
 
     -- custom requests
     , requestHandler (SMethod_CustomMethod (Proxy @Ladder.EvalAppMethodName)) $ \ide params ->
-        liftIO $ runVizHandlerM $ handleCustomVizRequest recorder (Proxy @Ladder.EvalAppMethodName) params ide $
+        liftIO $ runVizHandlerM $ withVizRequestContext recorder (Proxy @Ladder.EvalAppMethodName) params ide $
           \evalParams tcRes recentViz -> do
             let vizConfig = Ladder.getVizConfig . (.vizState) $ recentViz
             mEvalDeps <- liftIO $ runAction "l4/evalApp" ide $ use (AttachCallStack [vizConfig.moduleUri] GetLazyEvaluationDependencies) vizConfig.moduleUri
@@ -347,13 +347,13 @@ handlers recorder =
                 }
               Just (evalEnv, _) -> do
                 result <- MkVizHandler $ evalApp (evalEnv, tcRes.module') evalParams recentViz
-                logWith recorder Debug $ 
+                logWith recorder Debug $
                   LogReceivedCustomRequest evalParams.verDocId._uri
                   ("Eval result: " <> Text.show result)
                 pure result
 
     , requestHandler (SMethod_CustomMethod (Proxy @Ladder.InlineExprsMethodName)) $ \ide params ->
-        liftIO $ runVizHandlerM $ handleCustomVizRequest recorder (Proxy @Ladder.InlineExprsMethodName) (params :: MessageParams (Method_CustomMethod InlineExprsMethodName)) ide $ 
+        liftIO $ runVizHandlerM $ withVizRequestContext recorder (Proxy @Ladder.InlineExprsMethodName) (params :: MessageParams (Method_CustomMethod InlineExprsMethodName)) ide $
           \(_inlineExprsParams :: Ladder.InlineExprsRequestParams) _tcRes _recentViz -> undefined
     ]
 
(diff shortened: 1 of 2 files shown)


jl4-lsp/src/LSP/L4/Viz/CustomProtocol.hs
```diff
@@ -1,48 +1,78 @@
...
+-- l4/inlineExprs Request Params and Response Payload
+------------------------------------------------------
+
+-- | Payload / params for InlineExprsRequest
+data InlineExprsRequestParams = InlineExprsRequestParams
+  { uniques  :: [Int]
+  , verDocId :: LSP.VersionedTextDocumentIdentifier
+  }
   deriving stock (Eq, Show, Generic)
   deriving anyclass (FromJSON, ToJSON)
```
ym-han, line 65: The response payload gets refactored in the next PR

jl4-lsp/src/LSP/L4/Viz/CustomProtocol.hs
```diff
@@ -51,4 +81,18 @@ class KnownSymbol a => IsCustomMethod (a :: Symbol) where
...
+
+------------------------------------------------------
+--  IsLadderRequestParams typeclass
+------------------------------------------------------
+
+class (GHC.Records.HasField "verDocId" params VersionedTextDocumentIdentifier, 
+       FromJSON params) 
+      => IsLadderRequestParams params where
+  getVerDocId :: params -> VersionedTextDocumentIdentifier
+  getVerDocId = (.verDocId)
```
MangoIV, line 95: what does this actually do? 
isn't this just 
`getVerDocId = (.verDocIc)` (without class etc)

jl4-lsp/src/LSP/L4/Viz/CustomProtocol.hs
```diff
@@ -1,48 +1,78 @@
...
+    "$type" .= ("EvalAppResult" :: T.Text),
+    "value" .= val
+    ]
+
+instance FromJSON EvalAppResult where
+  parseJSON = withObject "EvalAppResult" $ \obj -> do
+    typeTag <- obj .: "$type"
+    if typeTag == ("EvalAppResult" :: T.Text)
+      then EvalAppResult <$> obj .: "value"
+      else fail $ "Expected $type field to be 'EvalAppResult' but got: " <> T.unpack typeTag
```
MangoIV, line 52: I'm guessing this happens because you cannot use the standard encoding on the other side?

MangoIV: where `the = haskell's`?

MangoIV (approved):
okay.

jl4-lsp/src/LSP/L4/Viz/CustomProtocol.hs
MangoIV: fyi - if you do 
`class Foo a => Bar a` then *all* `a` already need to implement `Foo`. So if the instances don't actually do anything, then you might as well just skip the class definition.

jl4-lsp/src/LSP/L4/Viz/CustomProtocol.hs
ym-han: ah it's just because, strictly speaking, the interface / schema that the frontend expects has a $type field:

```
export const EvalAppResult = Schema.Struct({
  $type: Schema.tag('EvalAppResult'),
  value: BoolValue,
})
```

If there's a cleaner or more idiomatic way to make sure that the encoded json has that $type field, I'm happy to switch it to that way.

It would also be reasonable to say that we shouldn't require a $type field for this in the first place

jl4-lsp/app/LSP/L4/Handlers.hs
```diff
@@ -45,7 +46,7 @@ import qualified LSP.Core.Shake as Shake
 import LSP.Core.Types.Diagnostics
 import LSP.Core.Types.Location
 import qualified LSP.L4.Viz.Ladder as Ladder
-import LSP.L4.Viz.CustomProtocol as Ladder
+import           LSP.L4.Viz.CustomProtocol as Ladder
```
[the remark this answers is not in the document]
ym-han: There wasn't a good reason for this --- I've changed it to the following

```
import qualified LSP.L4.Viz.CustomProtocol as Ladder
import LSP.L4.Viz.CustomProtocol (IsLadderRequestParams)
```

I'm assuming you would prefer this change, but let me know if you'd rather that it be something else. I don't have a good sense for what people tend to like or not like here.

jl4-lsp/src/LSP/L4/Viz/CustomProtocol.hs
ym-han: The overarching motivation for having a `IsLadderRequestParams` class, instead of, e.g., just using the specific constraints (fromJSON, hasField "verDocId"), is basically the same reason why one might want to work with an opaque newtype wrapper over a concrete monad transformer stack, along with operations on that opaque stack, instead of the concrete stack: I wanted to hide the implementational details re what exactly LadderRequestParams requires or amounts to.

I'm reasonably confident that this overarching motivation makes sense --- e.g., and imo, it already helps reduce cognitive load, and makes the code more readable. But I'm less confident about what the best way to implement or realize this motivation is. If there are cleaner or more idiomatic ways to do this kind of info-hiding, I'm happy to refactor accordingly.

And, yeah, there are intermediate solutions here: I could, e.g., keep the typeclass but do away with the `getVerDoc` function.

ym-han:
I'm going to go ahead and merge this in, but I can change the things I mentioned if you have further feedback

PR #450, opened 2025-05-20
Implement desugaring of infix binary functions

### LINKED ISSUE #382: desugar prefix operators to functions
in principle, this would allow us to re-implement operators.

fendor:
Desugars the arithmetic operators `+`, `-`, `*`, `/` and `modulo` to prefix functions:

* `__PLUS__`
* `__MINUS__`
* `__TIMES__`
* `__DIVIDE__`
* `__MODULO__`

Users can provide implementations for these functions to overload the infix binary operators for user-defined types.

Closes #382


fendor:
I still have to add overloading mechanisms for all the comparsion operators.

jl4-core/src/L4/Desugar.hs
```diff
@@ -0,0 +1,318 @@
...
+-- This works also nicely for parenthesis, the expression @(1 PLUS 2)@
+-- is then translated to:
+--
+-- @
+--   [["(", _, " PLUS "], [_, ")"]]
+-- @
+rewriteBinOpAnno :: Expr Name -> Expr Name -> Expr Name -> Desugar [Expr Name]
+rewriteBinOpAnno expr e1 e2 = do
+  let
+
```
MangoIV, line 279:
```suggestion
```

jl4-core/src/L4/Desugar.hs
```diff
@@ -0,0 +1,318 @@
...
+        , updateAnnoOf e2 (surroundWithCsn [] after)
+        ]
+    _slices -> internalError $ InternalAnnoRewritingError expr (HoleInfo 2 numberOfAnnoHoles)
+ where
+  annoPieces = (getAnno expr).payload
+
+  updateAnnoOf e f = e & annoOf %~ f
+
+  csnSlices = splitWhen (isJust . preview #_AnnoHole) annoPieces
+  numberOfAnnoHoles = length $ filter (isJust . preview #_AnnoHole) annoPieces
```
MangoIV, line 293: this all seems very fragile. The docs make sense thought and I don't have a better idea at this moment. Maybe let's talk about this.  

I also don't think this function warrants an additional dependency...

jl4-core/src/L4/Desugar.hs
```diff
@@ -0,0 +1,318 @@
...
+import L4.Print (prettyLayout)
+import L4.Syntax
+import qualified L4.TypeCheck.Environment as TypeCheck
+
+import qualified Base.Text as Text
+import Data.List.Split (splitWhen)
+import Optics
+
+newtype Desugar a = MkDesugar (Either DesugarError a)
+  deriving newtype (Functor, Applicative, Monad)
```
MangoIV, line 20: this seems a bit premature, but okay, I guess it doesn't hurt at all

jl4-core/src/L4/Desugar.hs
```diff
@@ -0,0 +1,318 @@
...
+
+data HoleInfo
+  = HoleInfo {expected :: !Int, got :: !Int}
```
MangoIV, line 23: this needs documentation, please :)

jl4-core/src/L4/Desugar.hs
MangoIV, line 1: This modules is giving me dejavus :P 

We really need to fix the traversal problem.

jl4-core/src/L4/Evaluate/ValueLazy.hs
```diff
@@ -46,6 +46,7 @@ data Value a =
   | ValObligation Environment MaybeEvaluated MaybeEvaluated (MaybeEvaluated' (Maybe RExpr)) RExpr (Maybe RExpr)
   | ValROp Environment RBinOp MaybeEvaluated MaybeEvaluated
   | ValUnaryBuiltinFun UnaryBuiltinFun
+  | ValBuiltinFun BuiltinFun
```
MangoIV, line 49: I think this should be consistently named, i.e. ValBinaryBuiltinFun

jl4-core/src/L4/Evaluate/Value.hs
```diff
@@ -27,7 +27,7 @@ instance NFData Value where
   rnf (ValString t)               = rnf t
   rnf (ValList vs)                = rnf vs
   rnf (ValClosure given expr env) = env `seq` rnf given `seq` rnf expr
-  rnf (ValUnaryBuiltinFun r)      = rnf r
```
MangoIV, line 30: doesn't this rely on the constructors of the `r` not having any lazy arguments? This could break in future without us noticing.

jl4-core/src/L4/Evaluate/ValueLazy.hs
```diff
@@ -78,7 +91,8 @@ instance NFData a => NFData (Value a) where
   rnf ValNil                      = ()
   rnf (ValCons r1 r2)             = rnf r1 `seq` rnf r2
   rnf (ValClosure given expr env) = env `seq` rnf given `seq` rnf expr
-  rnf (ValUnaryBuiltinFun r)      = rnf r
+  rnf (ValUnaryBuiltinFun r)      = r `seq` ()
+  rnf (ValBuiltinFun r)           = r `seq` ()
```
MangoIV, line 95: same as above, i.e. ``r `seq` ()`` relies on `r` being in normalform after a simple `seq`. that doesn't have to be the case - future refactorings may break us without noticing.

jl4-core/src/L4/EvaluateLazy/Machine.hs
```diff
@@ -990,6 +1017,25 @@ trueVal = ValConstructor TypeCheck.trueRef []
...
+--   ar <- ref na ad
+--   br <- ref nb bd
+--   pure $ ValClosure
+--     (MkGivenSig emptyAnno
+--       [ MkOptionallyTypedName emptyAnno ad Nothing
+--       , MkOptionallyTypedName emptyAnno bd Nothing
+--       ])
+--     (App emptyAnno TypeCheck.plusRef [App emptyAnno ar [], App emptyAnno br []])
+--     emptyEnvironment
+
```
MangoIV, line 1038: leftover?

jl4-core/src/L4/EvaluateLazy/Machine.hs
```diff
@@ -1101,13 +1147,35 @@ initialEnvironment = do
...
+      [ (ValBuiltinFun val, unique)
+      | (val, uniques) <-
+          [ (LtFun, TypeCheck.ltUniques)
+          , (LeqFun, TypeCheck.leqUniques)
+          , (GtFun, TypeCheck.gtUniques)
+          , (GeqFun, TypeCheck.geqUniques)
+          ]
+      , unique <- uniques
+      ]
+
```
MangoIV, line 1176: don't you just want to map the `Fun` over the uniques, then flatten ones and then allocate all? Seems to be a bit complicated for what it is... 

Perhaps you could also split up the traverse and the list.

jl4-core/src/L4/TypeCheck/Environment.hs
```diff
@@ -76,7 +82,59 @@ ceilingBuiltin = fun_ [number] number
...
+
+gtUniques :: [Unique]
+gtUniques = [gtNumberUnique,  gtStringUnique,  gtBoolUnique]
+
+geqUniques :: [Unique]
+geqUniques = [geqNumberUnique, geqStringUnique, geqBoolUnique]
+
+allComparisonUniques :: [Unique]
+allComparisonUniques =
+  ltUniques <> leqUniques <> gtUniques <> geqUniques
```
MangoIV, line 132: This doesn't appear to be used? Maybe I'm not seeing it though

jl4-core/src/L4/TypeCheck/Environment/TH.hs
```diff
@@ -10,14 +10,24 @@ import L4.Syntax as L4
...
 import L4.Parser.SrcSpan
 
-data Renamed = MkRenamed {orig :: String, chosen :: Maybe String}
+data Renamed = MkRenamed
+  { orig :: String
+  , chosen :: Maybe String
+  , variants :: Maybe [String]
+  -- ^ There cannot be more than 10 variants for the same name
+  -- at the moment.
+  -- This is an arbitrary cut-off, that is easily fixed.
```
MangoIV, line 19: this is a really unnecessary hack - there's no actual need to do this. Honestly, I would prefer if you had just given each of the variants a name, like 

leqString, etc. 

if this is starting to become a problem, we might need to look for a different abstraction, but this is not nice...

MangoIV (approved):
really cool! I would be happy if some of my review points were addressed. I'm also not happy about the additional work that the new traversal you added will cause - this is not only annoying when adding but also when extending.

jl4-core/src/L4/Evaluate/Value.hs
fendor: I thought you would prefer it, as you wrote it 6 lines above: https://github.com/smucclaw/l4-ide/blob/cec92d08a949ebec79280a36ceb65d9d2b1067e1/jl4-core/src/L4/Evaluate/ValueLazy.hs#L89

MangoIV:
unfortunately, since the LayoutPrinter uses the desguared version of the operators when printing a *value*, this breaks the printing. I think this needs to be fixed before we can merge this.

mengwong:
With any luck, this will give nice affordances for the money division @serrynaimo

PR #451, opened 2025-05-21
Viz: Add Bool lit variants to VizExpr / IRExpr (#369)

ym-han:


jl4-lsp/src/LSP/L4/Viz/Ladder.hs
```diff
@@ -415,17 +418,17 @@ inlineExpr :: VizState -> Int -> Decide Resolved -> Decide Resolved
 inlineExpr vs target = over decideBody $ transformOf (Optics.gplate @(Expr Resolved)) replace
   where
     replace :: Expr Resolved -> Expr Resolved
-    replace expr = 
-      if isRefOfTarget expr 
-      then 
+    replace expr =
+      if isRefOfTarget expr
+      then
       case lookupDefForInlining vs target of
```
ym-han, line 424: Thanks, editor formatter

ts-apps/decision-logic-visualizer/src/lib/eval/type.ts
```diff
@@ -17,9 +17,9 @@ export type UBoolVal = TrueVal | FalseVal | UnknownVal
 
 export function toUBoolVal(value: VE.UBoolValue): UBoolVal {
```
ym-han, line 18: refactor

jl4-lsp/src/LSP/L4/Viz/Ladder.hs
MangoIV: this is expected.

PR #454, opened 2025-05-21
Inlining expr UI

johsi-k:
I've decided to go with the '+' symbol for node expansion. Unlike directional indicators (e.g., arrows pointing up/down/left/right), '+' conveys a non-directional, spatially neutral action that aligns better with how expansion works in our graph: the layout adjusts globally rather than unfolding in-place along a fixed axis. 

Using a neutral symbol avoids misleading spatial cues and keeps the interaction consistent with common UI expectations around “revealing more.” 

Alternative symbols (e.g., arrows or ellipses) were considered but seem less intuitive for this context, as they typically imply directional growth or continuation rather than full structural unfolding.

### DIFF
--- a/ts-apps/decision-logic-visualizer/src/lib/displayers/flow/sf-custom-nodes/ubool-var.svelte
+++ b/ts-apps/decision-logic-visualizer/src/lib/displayers/flow/sf-custom-nodes/ubool-var.svelte
@@ -7,6 +7,7 @@ https://github.com/xyflow/xyflow/blob/migrate/svelte5/packages/svelte/src/lib/co
   import { useLadderEnv } from '$lib/ladder-env.js'
   import { UBoolVarLirNode } from '$lib/layout-ir/ladder-graph/ladder.svelte.js'
   import { Handle } from '@xyflow/svelte'
+  import * as Tooltip from '$lib/ui-primitives/tooltip/index.js'
 
   let { data }: UBoolVarDisplayerProps = $props()
 
@@ -29,19 +30,31 @@ TODO: Look into why this is the case --- are they not re-mounting the ubool-var 
   </div>
   {#if data.canInline}
     <div class="absolute bottom-1 right-1">
-      <button
-        class="px-0.5 text-[0.625rem] rounded border border-border bg-background hover:bg-accent hover:text-accent-foreground transition-colors duration-150"
-        onclick={() => {
-          console.log('inline lir id', data.originalLirId.toString())
+      <Tooltip.Provider>
+        <Tooltip.Root>
+          <Tooltip.Trigger>
+            <button
+              class="px-0.5 text-[0.625rem] rounded border border-border bg-background hover:bg-accent hover:text-accent-foreground transition-colors duration-150"
+              onclick={() => {
+                console.log('inline lir id', data.originalLirId.toString())
 
-          const uniq = (
-            data.context.get(data.originalLirId) as UBoolVarLirNode
-          ).getUnique(data.context)
-          l4Conn.inlineExprs([uniq], ladderEnv.getVersionedTextDocIdentifier())
-        }}
-      >
-        +
-      </button>
+                const uniq = (
+                  data.context.get(data.originalLirId) as UBoolVarLirNode
+                ).getUnique(data.context)
+                l4Conn.inlineExprs(
+                  [uniq],
+                  ladderEnv.getVersionedTextDocIdentifier()
+                )
+              }}
+            >
+              +
+            </button>
+          </Tooltip.Trigger>
+          <Tooltip.Content>
+            <p>Unfold to definition</p>
+          </Tooltip.Content>
+        </Tooltip.Root>
+      </Tooltip.Provider>
     </div>
   {/if}
   <Handle type="source" position={defaultSFHandlesInfo.sourcePosition} />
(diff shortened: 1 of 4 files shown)


ts-apps/decision-logic-visualizer/src/lib/displayers/flow/sf-custom-nodes/ubool-var.svelte
```diff
@@ -40,7 +40,7 @@ TODO: Look into why this is the case --- are they not re-mounting the ubool-var
           l4Conn.inlineExprs([uniq], ladderEnv.getVersionedTextDocIdentifier())
         }}
```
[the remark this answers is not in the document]
ym-han: I think there's a Tailwind utility for this (maybe `sr-only`) --- @johsi-k could you try looking into this perhaps?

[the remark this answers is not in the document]
johsi-k: Went with aria-label instead of sr-only for accessibility. Since + is a visual symbol without semantic meaning, aria-label is a more direct way to communicate intent (“Expand inline expression”) to screen readers. With sr-only, screen readers may announce both the visible symbol (“plus”) and the hidden label, which can lead to redundant or confusing output.

ym-han (approved):
Thanks for clarifying the comment

ts-apps/decision-logic-visualizer/src/lib/displayers/flow/sf-custom-nodes/ubool-var.svelte
```diff
@@ -39,8 +41,9 @@ TODO: Look into why this is the case --- are they not re-mounting the ubool-var
           ).getUnique(data.context)
           l4Conn.inlineExprs([uniq], ladderEnv.getVersionedTextDocIdentifier())
         }}
+        aria-label="Expand inline expression"
```
ym-han, line 44: Oh wait one more thing: Call it 'inline expression'

ym-han: Perhaps think a bit about how best to put it?

johsi-k: I agree it’s important to clarify what’s being expanded. I’ve gone with "Expand nested clause" to keep it concise and communicative for screen readers, while matching how legal users understand the interface. Rather than spelling out “with its definition,” which risks being overly verbose or ambiguous, this phrasing conveys both the structure (a nested clause) and the action (expansion) in a way that aligns with UI/UX and accessibility conventions.

ts-apps/decision-logic-visualizer/src/lib/displayers/flow/sf-custom-nodes/ubool-var.svelte
```diff
@@ -39,8 +42,9 @@ TODO: Look into why this is the case --- are they not re-mounting the ubool-var
           ).getUnique(data.context)
           l4Conn.inlineExprs([uniq], ladderEnv.getVersionedTextDocIdentifier())
         }}
+        aria-label="Expand nested clause"
```
ym-han, line 45: I'm not sure that it's really accurate to call this a 'nested clause' either?

ts-apps/decision-logic-visualizer/src/lib/displayers/flow/sf-custom-nodes/ubool-var.svelte
```diff
@@ -17,6 +17,9 @@ https://github.com/xyflow/xyflow/blob/migrate/svelte5/packages/svelte/src/lib/co
 <!-- Need to use data.bleh to maintain reactivity -- can't, e.g., do `const bleh = data.bleh` 
 TODO: Look into why this is the case --- are they not re-mounting the ubool-var component? 
 -->
+<!-- accessibility: Using aria-label instead of Tailwind’s sr-only because 
+ aria-label gives screen readers the intended action without also reading the “+” symbol 
+ 'Expand nested clause' is a concise, user-friendly hint tailored for legal drafters. -->
```
ym-han, line 22: No need to have `'Expand nested clause' is a concise, user-friendly hint tailored for legal drafters.` in the comment

ym-han (changes requested):
Thanks for clarifying the comment

ts-apps/decision-logic-visualizer/src/lib/displayers/flow/sf-custom-nodes/ubool-var.svelte
johsi-k: I can see how “replace with definition” fits the underlying behavior, especially since it’s currently irreversible. Just to clarify what I meant about the destructive or editing action — I was thinking more along the lines of overwriting text or removing something permanently. In this case, the action is read-only and just reveals more detail — there’s no editing or input from the user. For screen reader users especially, verbs that suggest intent (like “expand” or “reveal”) might be a better fit, since they don’t have the visual context of what’s changing.

Even if it’s technically a one-way substitution, I wonder if it might be clearer to describe what the user experiences — a non-destructive reveal of a definition — rather than how the system implements it under the hood.

Maybe a middle ground could be something like “Unfold to definition” or “Expand term into definition”?

ym-han: “Unfold to definition” and “Expand term into definition” (as well as "Expand with definition") seem fine to me. Feel free to pick whichever you like

johsi-k: Thanks! I've added an "Unfold to definition" tooltip and aria-label for now

ym-han (approved):
Nice, looks good!
image

PR #465, opened 2025-05-23
Viz: Experiment with adding a non-color-based visual indicator of the uboolvalue

ym-han:
image


johsi-k (commented):
Thanks for taking a stab at this! Just a few thoughts on how we might refine the truth value indicators.

One possible refinement: rather than using just icons (✓ / ✗) or placeholder text like `"TRUE"`, we could try overlaying small, explicit text tags like `TRUE`, `FALSE`, or `UNKNOWN`. This might help make the logic status clearer at a glance — especially as nodes become more variable in size, carry longer bits of text, and as we add interaction layers like node selection and path-building.

This idea is less about replacing what we have and more about adding semantic clarity, especially in contexts where visual space is quite loaded.


What this might look like:

```
┌──────────────────────────────────────────────────────────┐
│ person p `is born in a qualifying territory after...`    │
│                                                          │
│                                                   [TRUE] │
└──────────────────────────────────────────────────────────┘
```

* Text overlays (e.g. `TRUE`, `FALSE`, `UNKNOWN`) in a consistent position — probably top or bottom right
* Small caps or monospaced style to signal logical status
* Optional subtle tag-like styling:

  * TRUE: light green background with green border
  * FALSE: light red with red border
  * UNKNOWN: neutral gray
* The idea is to support legibility without relying entirely on color — so color serves more as a gentle cue than a dependency



Why it might help:

* Text conveys meaning directly — less interpretation needed than symbols/icons
* Consistent positioning avoids ambiguity in larger or differently sized nodes
* Shouldn’t interfere with future interaction modes (e.g. clicking, shift-selecting, highlighting paths)
* Could work well with assistive tech if rendered semantically

ym-han:
> Thanks for taking a stab at this! Just a few thoughts on how we might refine the truth value indicators.
> 
> One possible refinement: rather than using just icons (✓ / ✗) or placeholder text like `"TRUE"`, we could try overlaying small, explicit text tags like `TRUE`, `FALSE`, or `UNKNOWN`. This might help make the logic status clearer at a glance — especially as nodes become more variable in size, carry longer bits of text, and as we add interaction layers like node selection and path-building.
> 
> This idea is less about replacing what we have and more about adding semantic clarity, especially in contexts where visual space is quite loaded.
> 
> What this might look like:
> 
> ```
> ┌──────────────────────────────────────────────────────────┐
> │ person p `is born in a qualifying territory after...`    │
> │                                                          │
> │                                                   [TRUE] │
> └──────────────────────────────────────────────────────────┘
> ```
> 
> * Text overlays (e.g. `TRUE`, `FALSE`, `UNKNOWN`) in a consistent position — probably top or bottom right
> * Small caps or monospaced style to signal logical status
> * Optional subtle tag-like styling:
>   
>   * TRUE: light green background with green border
>   * FALSE: light red with red border
>   * UNKNOWN: neutral gray
> * The idea is to support legibility without relying entirely on color — so color serves more as a gentle cue than a dependency
> 
> Why it might help:
> 
> * Text conveys meaning directly — less interpretation needed than symbols/icons
> * Consistent positioning avoids ambiguity in larger or differently sized nodes
> * Shouldn’t interfere with future interaction modes (e.g. clicking, shift-selecting, highlighting paths)
> * Could work well with assistive tech if rendered semantically

Feel free to experiment with this. I didn't go for this approach because I wanted something concise and not too text-heavy, but it's definitely worth experimenting with this if you want

ym-han:
(The value indicator is buggy for App args, but that will be fixed in another PR)

PR #466, opened 2025-05-23
jl4-cli: distinguish success vs failure runs via exitcode

mengwong:
not sure if this is the correct way to deal with multiple recorders, but it works.

upon success, exits 0

upon failure, exits nonzero

upon success, also refrains from printing the reformatted input to output, unless called with `-v`

### DIFF
--- a/jl4/app/Main.hs
+++ b/jl4/app/Main.hs
@@ -3,7 +3,7 @@ module Main where
 import Base (NonEmpty, for_)
 import Base.Text (Text)
 import Data.List.NonEmpty (some1)
-import Options.Applicative (fullDesc, header, helper, info, metavar, strArgument, help, short, long, switch)
+import Options.Applicative (fullDesc, header, footer, helper, info, metavar, strArgument, help, short, long, switch, progDesc)
 import qualified Options.Applicative as Options
 import System.Directory (getCurrentDirectory)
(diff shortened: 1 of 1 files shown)


fendor (commented):
Why not `oneshotL4ActionAndErrors`? Doesn't it do what you want?

mengwong:
Main sets up a recorder in a different way than in Oneshot, and the AndErrors variant doesn't give a way to use that recorder.

jl4/app/Main.hs
```diff
@@ -41,13 +47,28 @@ main = do
...
+
+data Options = MkOptions
+  { files :: NonEmpty FilePath
+  , verbose :: Bool }
 
 optionsDescription :: Options.Parser Options
-optionsDescription = MkOptions <$> some1 (strArgument (metavar "FILES..."))
+optionsDescription = MkOptions
+  <$> some1 (strArgument (metavar "FILES..."))
+  <*> switch (long "verbose" <> short 'v' <> help "Enable verbose output")
```
MangoIV, line 70: I think if you add a `help` option, then that option should say what it does, exactly, i.e. it also changes the way the exit codes behave.

MangoIV (approved):
I think this is okay.

PR #468, opened 2025-05-23
Implement ":" as a synonym for "IS"

### LINKED ISSUE #461: if `:` (colon) is not already used as a token, have it be a synonym for IS
this allows conventional record syntax

fendor:
In particular, in the following cases:

* Function Arguments
* Type Arguments
* Record field definitions

Articles are valid with `:`, e.g.
```
GIVEN a: A BOOLEAN
foo a MEANS a AND a
```

The token ":" is always hidden and will not be displayed in error messages to the user.

Closes #461


MangoIV (approved):
nice and easy, lovely

PR #469, opened 2025-05-23
more sophisticated highlight paths

ym-han:


ts-apps/decision-logic-visualizer/src/lib/layout-ir/ladder-graph/ladder.svelte.ts
```diff
@@ -1122,75 +1186,3 @@ export function augmentEdgesWithExplanatoryLabel(
     )
   })
 }
-
```
ym-han, line 1125: unrelated refactor

PR #494, opened 2025-05-28
Finish implementing the more sophisticated select / highlight paths

ym-han:


ts-apps/decision-logic-visualizer/src/style.css
```diff
@@ -290,3 +281,21 @@ body {
     font-style: normal;
   }
 }
+
+/* Override border properties when highlighted */
+.highlighted-ladder-node {
```
ym-han, line 286: not working yet

PR #495, opened 2025-05-29
tgorissen/indentation magic

mengwong:
improve VS Code UX

- **indentation magic: tweak package.json**
- **indentation magic from commit 461de88**

### DIFF
--- a/ts-apps/vscode/language-configuration.json
+++ b/ts-apps/vscode/language-configuration.json
@@ -1,0 +1,79 @@
+{
+  "comments": {
+    "lineComment": "--",
+    "blockComment": ["{-", "-}"]
+  },
+  "brackets": [
+    ["[", "]"],
+    ["(", ")"]
+  ],
+  "autoClosingPairs": [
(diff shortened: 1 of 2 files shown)


mengwong:
I believe this was discussed during the call, and the decision was to simplify a bit? Not 100% sure, I only overheard some of the discussion. In any case the CI seems to be failing, but I'm not sure if that's a transient error or not. @serrynaimo

mengwong:
hey @kharus any idea why these tests might be failing? Do we need to just try again?

image

ym-han:
> I believe this was discussed during the call, and the decision was to simplify a bit? Not 100% sure, I only overheard some of the discussion. In any case the CI seems to be failing, but I'm not sure if that's a transient error or not. @serrynaimo

It's fine, don't worry about it. I was saying that it might be good to change the config in incremental steps to be safe, but it's probably fine.

mengwong:
obsoleted by some other PR, i think

PR #516, opened 2025-06-02
move recent experiments into main. This should not affect anything.

mengwong:


mengwong:
Unless there are any objections I will go ahead and merge this.

serrynaimo (approved):
Should be fine

PR #521, opened 2025-06-03
Revert "[feat] add support for arbitrary literal patterns"

MangoIV:
Reverts smucclaw/l4-ide#512


MangoIV:
will be replaced by Foo (EXACTLY 3 + 3)

PR #545, opened 2025-06-09
jl4-web: Experiment with using monaco-error-lens with jl4-web to make diagnostics more visually prominent

ym-han:
image


ym-han:
I'm going to merge this in for now; it's easy to revert it if we don't like it

Issue #549, opened 2025-06-10
Replace the stock example in jl4-web with something that shows off more language and visualizer features


ym-han:
- [ ] functiona
- [ ] datatypes 
- [ ] assumes 
- [ ] syntax 
- [ ] contract syntax 
- [ ] `#EVAL`

ym-han:
Also related to #381

mengwong:
Resolved by commit `e6923bdd` ('Replace hardcoded example in jl4-web with test-validated examples'), which swapped the single hardcoded example for a curated, test-validated rotation (`jl4/examples/legal/*.l4`) that between them cover functions, datatypes, `ASSUME`, contract syntax (`PARTY`/`MUST`/`WITHIN`/`LEST`), and all `#EVAL` forms (see `directive-showcase.l4` and `promissory-note.l4` in particular). Closing — thanks for flagging this.

PR #562, opened 2025-06-12
Replace hardcoded example in jl4-web with test-validated examples

johsi-k:
This partly addresses #549. 

The web UI currently displays a hardcoded BNA example in +page.svelte, which isn’t covered by our test infrastructure. At the same time, we already maintain a set of l4 examples in jl4/examples/legal/ that are automatically validated during testing.

This PR replaces the hardcoded example with a setup that loads the validated examples instead. It ensures that what we show in the interface is always aligned with our tested examples, and simplifies contributor workflow.

image

@MangoIV: I've modified `nix/jl4-web/package.nix` to fix an issue where the nix build was failing because the `copy-examples.ts` script couldn't find the `jl4/examples/legal` directory. The script expects the full directory structure to be present, but our source filtering was only including the files.

The change I've made uses `lib.cleanSourceWith` with a filter that:
- preserves directory structure by including all directories (`type == "directory"`)
- selectively includes only the files we need (TypeScript files, package files, and legal examples)

While this works, I'm not sure this is the most idiomatic approach. I'd appreciate your help on:
- whether there's a more elegant way to handle directory structure preservation
- if we could achieve the same result with less complex filtering logic
- any potential performance implications of including all directories

Thanks!


MangoIV:
@johsi-k I would like to ask two questions:
- is this PR so far ready?
- I am not sure about the new "side bar" taking up too much space - could we e.g. split the visualizer horizontally and use a drop down? or did you consider this and it didn't turn out well / wasn't discoverable enough. 
thanks in advance!

ym-han:
Re whether to make the copied / generated `examples` a subdir of `static` or `build`: just google for conventions re static vs build in svelte projects; hopefully those conventions will be enough

johsi-k:
@MangoIV This PR is ready for review. I’ll hold off on any further UI changes so this doesn’t fall too far behind main, those can go into a separate PR.

MangoIV (approved):

PR #573, opened 2025-06-12
Lexer refactoring

### LINKED ISSUE #572: improve ergonomics of lexer
since we add a lot of new constructs, I think the ergonomics of the lexer could be improved:
- offer exactly one map between tokens and their names
- categorize tokens as part of a nesting of the token datatype.

MangoIV:
#571 should be closed first
Closes #572

### DIFF
--- a/jl4-core/src/L4/Lexer.hs
+++ b/jl4-core/src/L4/Lexer.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE PatternSynonyms #-}
 {-# OPTIONS_GHC -Wno-orphans #-}
 module L4.Lexer where
 
@@ -49,7 +50,7 @@ data AnnoType
   deriving stock (Eq, Ord, Show, Generic)
   deriving anyclass (ToExpr, NFData)
 
-data DirectiveType
+data TDirectives
   = TStrictEvalDirective
   | TLazyEvalDirective
   | TCheckDirective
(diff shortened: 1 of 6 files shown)


jl4/examples/legal/tests/imaginary-alcohol-just-propositions.ep.golden
```diff
@@ -0,0 +1,97 @@
+{-
```
ym-han, line 1: Sorry about forgetting to add these!

jl4-core/src/L4/Parser.hs
```diff
@@ -1131,6 +1132,23 @@ ifthenelse = do
...
+        (,)
+          <$> do
+             annoLexeme (spacedToken_ TKIf)
+              *> annoHole (indentedExpr current)
+          <*> do
+             annoLexeme (spacedToken_ TKThen)
+              *> annoHole (indentedExpr current)
+      <*  annoLexeme (spacedToken_ TKOtherwise)
+      <*> annoHole (indentedExpr current)
+
```
fendor: The use of `many` feels weird, it should likely still be `indented` in some way, otherwise the following is accepted:

image

Which looks like a bug to me

fendor (changes requested):
LGTM, one issue

jl4-core/src/L4/Lexer.hs
```diff
@@ -855,129 +914,38 @@ displayPosToken :: PosToken -> Text
 displayPosToken (MkPosToken _r tt) =
   displayTokenType tt
 
+inversedCompleteLookup :: (Eq a, Show a) => a -> Map c a -> c
+inversedCompleteLookup k = fromJust . (\case {Nothing -> error ("error while looking up: " <> show k); a -> a}) . lookup k . map (\(a, b) -> (b, a)) . Map.toList
```
fendor, line 918: :scream: 

more line breaks, more comments, more info in the impossible `error` case, please!

fendor (approved):
LGTM, one nitpick

jl4-core/src/L4/Lexer.hs
MangoIV: oh also this is a leftover, it has fromJust already :P

PR #588, opened 2025-06-20
allow to load jl4 programs from their uuid into the decision service

MangoIV:
- allow to GET and PUT functions from a document with a certain UUID by specifying the function like `<uuid>:<function-name>` 
- the function needs to have a given signature 
- the uuids are not specified when `GET`ting all functions. This is a security feature.

### DIFF
--- a/jl4-decision-service/src/Examples.hs
+++ b/jl4-decision-service/src/Examples.hs
@@ -103,7 +103,7 @@ personQualifiesFunction = do
                 |]
         , parameters =
             MkParameters
-              { parameters =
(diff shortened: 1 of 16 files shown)


jl4-decision-service/src/Examples.hs
```diff
@@ -154,7 +154,7 @@ rodentsAndVerminFunction = do
                   ]
             in
               MkParameters
-                { parameters = params
+                { parameterMap = params
```
MangoIV, line 157: ambiguous record updates :/

PR #595, opened 2025-06-24
Add an ASSERT directive.

kosmikus:
See #592 .

### DIFF
--- a/jl4-core/src/L4/EvaluateLazy.hs
+++ b/jl4-core/src/L4/EvaluateLazy.hs
@@ -1,6 +1,7 @@
 {-# LANGUAGE GADTs #-}
 module L4.EvaluateLazy
 ( EvalDirectiveResult (..)
+, EvalDirectiveValue(..)
 , execEvalModuleWithEnv
 , execEvalExprInContextOfModule
 , prettyEvalException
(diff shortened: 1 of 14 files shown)


kosmikus:
The idea here is that `#ASSERT` takes a `BOOLEAN`. It's an info-diagnostic on success, and an error-diagnostic on failure.

jl4-core/src/L4/EvaluateLazy.hs
```diff
@@ -225,8 +226,16 @@ nfDirective (MkEvalDirective r traced expr env) = do
...
         nf whnf
-  let finalTrace = postprocessTrace <$> mt
-  pure (MkEvalDirectiveResult r v finalTrace)
+  let
+    finalTrace = postprocessTrace <$> mt
+    v' =
+      if isAssert
+        then Assertion
+          case v of
+            Right (MkNF nf') | Just True <- boolView nf' -> True
```
MangoIV, line 235: you could have replaced this with `Right (MkNF (ValBool True))`, theoretically..

jl4-core/src/L4/TypeCheckNew.hs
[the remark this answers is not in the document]
kosmikus: Yes, my mistake. Don't know why this happened.

jl4-core/src/L4/EvaluateLazy.hs
MangoIV: Plausible. Nvm then.

PR #596, opened 2025-06-24
Add option to hide Examples Pane with `no-examples` parameter

serrynaimo:


serrynaimo:
It is. There is more work required to make the sidebar something useful. It gives a very half-hearted impression right now where it doesn't quite know why it is there, just linking to a bunch of random examples. Maybe I can create an .md file to render in the sidebar for context later

johsi-k (changes requested):
This looks great, but before approving, I wanted to flag a subtle but important issue with the current 3-pane, 2-handle resizable layout (Example Selector | Editor | Visualizer):
When toggling the Example Selector pane (show/hide), the resizing handles can start behaving unpredictably: sometimes dragging the left handle doesn’t resize the expected panes, or stops working altogether. After toggling, the handles can end up controlling the wrong panes, or one handle may stop working. This is a known class of bug in resizable panel libraries when the number of panes/handles changes dynamically at runtime.

Why this happens:
The resizable component internally maps handles to panes at initial render. When we conditionally render the Example Selector as a resizable pane and toggle it, the internal mapping can get out of sync with the DOM, leading to these bugs.

Suggested minimal fix:
If we don’t need the Example Selector to be resizable, I recommend rendering it as a plain `<div>` instead of a `<Resizable.Pane>`, and removing the left `<Resizable.Handle>`. This way, the resizable logic only manages the Editor and Visualizer, and the Example Selector is simply shown/hidden as a static sidebar. This should eliminate the handle confusion and make the resizing behavior robust again.

The previous overlay approach (2 panes, 1 handle, Example Selector as a floating overlay) was robust for this reason, but I’m not insisting we revert to that if you prefer the sidebar as a pane. I think the key is to avoid dynamic pane/handle changes in the resizable group.

serrynaimo:
Great point! I moved the sidebar into a div by itself! Thanks!

PR #601, opened 2025-06-25
a collection of small changes

### LINKED ISSUE #589: explain how the backend services work together.

### LINKED ISSUE #591: synonymous MUST MAY DO
I remember we had a discussion about just giving the object-level language alternative options for MUST / MAY / DO,

possibly with syntactic sugar for HENCE / LEST defaults;

the actual "is it true that the party must always x" would be shifted to the property-level language.

Sorry I forget the status of this discussion, creating an issue to track it, in case we don't already have an issue for it.

### LINKED ISSUE #597: web: share button should update the location bar
Image

after clicking the share button, consider doing a `history.replaceState` to update the bar with the same URL that was copied to clipboard.

### LINKED ISSUE #600: Add "empty document" to the example files in the web ide

MangoIV:
- allow may and do in place of must (closes #591) 
- add empty document to examples (closes #600)
- push the URL to the history after sharing (closes #597)
- explain how the services play together (closes #589)

### DIFF
--- a/Dev.md
+++ b/Dev.md
@@ -76,7 +76,17 @@ Open folder [./jl4/examples](./jl4/examples) to see the Language Server in actio
 
 Click on "visualize" to see a rendering of the decision logic.
 
-## Running the web id locally
+## Running the web IDE locally
+
+The general overview is as follows: 
+
+- the webpage is served as a javascript bundle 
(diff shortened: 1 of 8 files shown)


fendor (approved):
LGTM

PR #607, opened 2025-06-27
[feat] document contracts as implemented in here

MangoIV:


kosmikus (approved):
All good as far as I can see.

PR #648, opened 2025-11-27
Improve error message when printing unapplied functions

mengwong:
When evaluating an unapplied function (e.g., `#EVAL add` where `add` takes two arguments), the output now shows a helpful message: "<function: can't print a function -- are you calling it with enough arguments?>"

This helps users understand they need to provide all required arguments to a function before it can be evaluated to a printable value.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mengwong:
This is actually not that useful. Closing.

Issue #635, opened 2025-09-29
Critical L4 decision service improvements for active service


serrynaimo:
Throughout testing I found the following limitations to be show-stoppers when it comes to deploying the decision service in larger production environments. In order of importance:

- [ ] A syntax to make a function available as evaluation API. My suggestion `EXPORT \`identifier name\``. That implies that multiple functions can be made available and it could be useful to allow `EXPORT DEFAULT \`identifier name\`` per file as well. § scoping should still be considered.
- [ ] The decision service should ignore all #EVAL #ASSERT #CHECK etc uses in a document. Those should be exclusively considered for IDE/other LSP use. Also to improve performance of decision service.
- [ ] IMPORTs must be working properly with the decision service for loaded files supporting only downstream folder structures. EXPORTs in imported files not loaded directly into the decision service should be available as well but somehow import scoped. (not to confuse with § scoping?)
- [ ] Performance must be improved a lot. I think ignoring # keywords will go a long way and also not compiling for debug, but it really needs to run super quick and with low resource usage even for thousands of loaded files.
- [ ] Load and unload L4 files without service restart
- [ ] Return decision trace only with additional header setting (for performance, e.g. -x-jl4-trace: "full|partial")

That should be good enough for now. Deeper rabbit holes like authentication we go into later ;)

mengwong:
Re-audited the full checklist against `origin/unstable` (2026-07-07). Every item is resolved except the one already explicitly deferred by this issue's own checklist:

- **Trace header** — done (`131dd4a0`).
- **Ignore `#EVAL`/`#ASSERT`/`#CHECK` in decision service** — done (`00ea68b0`).
- **Nested-object YAML support** — done (`00ea68b0`).
- **Boolean Minimization** — intentionally deferred to #638 ("for consideration"), not blocking this issue. (Side note: `jl4-query-plan`'s `L4.Decision.BooleanDecisionQuery`/`L4.Decision.QueryPlan`, exposed via `jl4-service`'s `POST .../query-plan` endpoint, already implements a real BDD engine with elicitation ordering / "don't care" pruning that substantially overlaps with #638's ask — worth a look over there, but out of scope for closing this issue.)
- **Load/unload without restart** — done (REST CRUD `POST/PUT/DELETE /functions/:name` + UUID-based loading, `e330a8a3`).
- **EXPORT syntax** (`@export`) — done, `specs/done/EXPORT-SYNTAX-SPEC.md`, all 4 phases implemented.
[...]

PR #651, opened 2025-11-29
some priorities associated with #635

mengwong:
Trying to spec and implement some of the bullet points identified under #635.

The use of JSONDECODE is expected to lead to conflict if we soon change the Maybe return type to an Either.


mengwong:
@codex review

PR #702, opened 2025-12-18
Document mixfix/postfix fix

mengwong:
## Summary
- clarify in the mixfix spec that backticks are purely lexical and reference the new postfix regression programs
- capture the shipped safety net in the implementation plan plus status docs
- update the postfix-with-variables issue and historical notes with the new behavior

## Testing
- npm run format
- cabal test all


mengwong:
obsoleted by #704

PR #739, opened 2026-01-09
Add VSCode Extension workflows and ability to create binaries for multiple platforms

serrynaimo:
@mengwong For your review. This attempts to build the two binaries for Mac ARM and Windows and bundles them into the extension.

### DIFF
--- a/ts-apps/vscode/README.md
+++ b/ts-apps/vscode/README.md
@@ -69,6 +69,7 @@ Alternatively, you can specify the location of the binary via `.vscode/settings.
 #### Binary Resolution Order
 
 The extension looks for the language server in this order:
+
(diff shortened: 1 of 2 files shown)


mengwong:
This opens the door to a larger discussion – I think our userbase may segment into:

- people who run the LSP server locally for speed and privacy

- people who run against our cloud LSP server using the wss protocol, for convenience; they may not be allowed to install a local binary as part of the VS Code extension, yet they may be less concerned about the privacy implications of their source code traveling over the net.

In the second instance, freemium pricing may apply, for the sake of sustainability -- if we're serving LSP at scale we need to cover costs.

So we should look into things like Stripe's API billing capabilities.

PR #760, opened 2026-01-12
feat(lint): Add AND/OR same-column warning

### LINKED ISSUE #758: LSP linter should warn if AND and OR are seen at the same level of indentation
we do have operator precedence, but humans make mistakes, so let's give it a yellow background or yellow squiggle. It's OK to have a load of ANDs (`...`) at the same indentation level; it's OK to have a load of OR (`..`) at the same level; but if the same level mixes AND/OR then it's PROBABLY a code smell, and the LSP plugin should highlight that.

mengwong:
## Summary
- Adds LSP linter warning when AND and OR operators appear at same indentation level on different lines
- New module L4.Lint.AndOrDepth extracts keyword positions from Anno CSN payload for accurate detection
- Warns about precedence confusion patterns

## Test plan
- [x] Test file and-or-simple-test.l4 with basic positive/negative cases
- [x] Test file and-or-depth-warning.l4 with comprehensive edge cases
- [x] All 726 tests pass
- [ ] Manual verification in VS Code extension

Fixes #758

Generated with Claude Code via Happy

### DIFF
--- a/jl4-core/jl4-core.cabal
+++ b/jl4-core/jl4-core.cabal
@@ -93,6 +93,7 @@ library
     L4.FindDefinition
     L4.JsonSchema
     L4.Lexer
+    L4.Lint.AndOrDepth
     L4.Mixfix
     L4.Names
     L4.Nlg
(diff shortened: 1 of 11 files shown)


mengwong:
Addressed Codex review: The limited recursion scope is intentional. The linter checks within a single boolean construct only - nested constructs (IfThenElse, Consider, Where, LetIn, etc.) are separate logical scopes, and each DECIDE body is checked independently. Added clarifying documentation to the code in commit f9ae9b73.

